PR #1, opened 2021-07-13
Add the PDF Image loader

lolipopshock:


lolipopshock:
After some reconsideration, I think it would probably be better if the `load_image` function exist in `Document`. When calling `doc.images`, it will identify whether the document image is loaded - If not, it will run the commands to have it loaded. The reason is very simple - efficiency. It's an expensive operation and we only want to have it when needed. 

But there are definitely some drawbacks. And in order to implement this function, we need some modifications for the `Document` class - for example, having the page size and original file path stored.

lolipopshock:
1. The images are loaded by default, and can be disabled via `parser.parse(..., load_images=False)`. 
2. Will add readme and tests later today.

kyleclo (approved):
,lgtm

PR #4, opened 2021-07-16
Scaffolding the new document classes and elements

lolipopshock:


lolipopshock:
Major design concepts: 

1. For each document, there will be two types of source data/representation: 
    1. `DocumentSymbols`
    2. `DocumentImages`

    They are stored as a list of elements for each page. 
2. We have two major types of document annotations: 
    1. `Span` 1D structure
    2. `Box` 2D structure 
    
    , and they derive the corresponding compound annotations classes: 
    
    1. `SpanGroup` for a list of `Span`s
    2. `BoxGroup` for a list of `Box`es

    But all these classes are only for storing "coordinates", and no textual information is stored. 

3. The `DocSpan` and `DocBox` are the actual classes we will be using for representing the Span or box-based text annotation over the documents.

PR #7, opened 2021-08-03
Implement `annotate` for DocSpanGroup

lolipopshock:


lolipopshock:
I am not sure about how DocBoxGroup will be used in practice, so only implemented `annotate` for `DocSpanGroup`.

lolipopshock:
And somehow I don't the current APIs for docspangroup is "elegant" -- for example, if you need to get the "start" and "end" for them, you have to:
```python
word = DocSpanGroup(xxx)
word.spans[0].start, word.spans[0].end
```

Can we just build a proxy function to make ppl's life easier? 

```python
@dataclass
class DocSpanGroup(DocumentAnnotation):
    
    @property
    def start(self):
        return xxx
    
    @property
    def end(self):
        return xxx 
```

kyleclo (approved):

PR #8, opened 2021-08-03
Kylel/2021 08 02/file reorg

kyleclo:
Mostly just moving files around.  Previously, everything was in 2 files `document.py` and `document_elements.py`.  Now it's split into:

```
# primary house for data
|-- document.py     
|-- document_elements.py

# base structures, including ImageGroup and SpanGroup
|-- image.py
|-- span.py
|-- box.py

# Document-aware wrapper around groups like SpanGroup.  Also their respective indices.
|-- annotation.py
```

Also include some bugfixes like wrong variable names.  Added some comments


mmda/types/document_elements.py
```diff
@@ -1,11 +1,16 @@
+"""
+
+
+"""
+
+# TODO[kylel] not sure this class needs to exist; seems extra boilerplate for no benefit
```
lolipopshock, line 6: I think it would help:
1. for extensibility for future updates that stores token-level bbox
2. for indexing tokens across pages: now we support something like `tokens[page_id, start:end]`

But yeah.. could be removed if you think that's more straightforward

lolipopshock:
LGTM! Thanks!

PR #9, opened 2021-08-03
major refactor - doc, annotations, from/to_json

kyleclo:
Remove the page-based indexing for symbols 
Consolidate DocSpanGroup and Spangroup 
Remove upgrading from Annotate & only takes consolidated SpanGroup
from_json in Doc use the annotate to create indexers 
Add from_json() to each base unit below a Doc


mmda/types/annotation.py
```diff
@@ -95,27 +116,26 @@ def to_json(self) -> Dict:
...
-class DocSpanGroupIndexer(Indexer):
 
-    num_pages: int
+@dataclass
+class SpanGroupIndexer(Indexer):
 
-    # TODO[kylel] -- explain why?
     def __post_init__(self):
-        self._index: List[IntervalTree] = [IntervalTree() for _ in range(self.num_pages)]
+        self._index: IntervalTree = IntervalTree()
```
lolipopshock, line 129: How about having `_index` in the dataclass init?  
```python
class SpanGroupIndexer(Indexer):
    _index: IntervalTree = IntervalTree()

```

mmda/types/box.py
```diff
@@ -18,16 +18,11 @@ class Box:
     h: float
     page: int
 
+    # TODO[kylel]: make this more minimal
     def to_json(self) -> Dict:
         return dict(l=self.l, t=self.t, w=self.w, h=self.h, page=self.page)
```
lolipopshock, line 23: how about `vars(self)`?

mmda/types/document.py
```diff
@@ -12,96 +12,140 @@
...
-        self._register_field(field_name)
-        
-        for annotation in field_annotations:
-            annotation = annotation.annotate(self)
-            
-        setattr(self, field_name, field_annotations)
-
-    def annotate(self, **annotations: List[Annotation]):
+    # TODO: extend implementation to support DocBoxGroup
+    def find_overlapping(self, query: Annotation, field_name: str) -> List[Annotation]:
```
lolipopshock, line 39: why `find_overlapping` rather than `find`? because probably `find` will be used for other purposes?

mmda/types/annotation.py
```diff
@@ -30,63 +30,84 @@ def to_json(self) -> Dict:
...
     @abstractmethod
-    def annotate(self, doc: "Document"):
-        """Annotate the object itself on a specific document.
-        It will associate the annotations with the document symbols.
-        """
+    @classmethod
+    def from_json(cls, annotation_dict: Dict) -> "Annotation":
+        pass
+
+    def attach_doc(self, doc: "Document") -> None:
```
lolipopshock, line 37: I like it!

mmda/types/document.py
```diff
@@ -12,96 +12,140 @@
...
 
+    @classmethod
+    def from_json(cls, doc_dict: Dict) -> "Document":
+        fields = doc_dict.keys()        # TODO[kylel]: this modifies the referenced dict, not copy
+
+        # 1) instantiate basic Document
+        symbols = fields.pop(Symbols)
+        images_dict = doc_dict.pop(Images, None)
+        if images_dict:
+            raise NotImplementedError(f'We havent figured out how to instantiate Images from JSON yet')
```
lolipopshock, line 126:
```python
from mmda.types.image import Image as DocImage
images = [DocImage.frombase64(image_str) for image_str in images_dict]
```

lolipopshock:
LGTM!

mmda/types/annotation.py
kyleclo: done

mmda/types/box.py
kyleclo: defer to later

mmda/types/document.py
kyleclo: just for transparency sake. that this is an `overlapping` cal.  the reason for this is cuz of some discussion we had where how `find()` operates really is specific to this overlapping nature

mmda/types/document.py
kyleclo: done

PR #10, opened 2021-08-07
Kylel/2021 08 07/sscraper

kyleclo:
1. shortened serialization of boxes from a Dict to a List
2. modified BaseParser parse() arguments to be a bit simpler
3. update SScraper implementation to use SpanGroups instead of Span
4. some python dataclasses bugfixes
5. add new method to Box to compute a Union box
6. add common field names to types.names
7. update sscraper test to match new serialization method


mmda/parsers/parser.py
```diff
@@ -12,21 +12,20 @@
...
-        self, infile: str, outdir: Optional[str] = None, outfname: Optional[str] = None, load_images=True
-    ) -> Union[str, Document]:
+    def parse(self, input_pdf_path: str, output_json_path: Optional[str] = None,
+              tempdir: Optional[str] = None, load_images: bool = False) -> Union[str, Document]:
         """This is the main entrance point for using the PDF parsers. For a
         given PDF file, this method will return a Document object.
         """
 
+    # TODO[kylel] - this doesnt quite work; convert_from_path() doesnt use our Image w patches; it uses PIL.Image
+    # TODO[kylel] - serialization?
```
lolipopshock, line 27: Yes, but fundamentally the output type is just an inherited class based on `PIL.Image` that is designed for handling loading images of different types. I had this initial confusion as you did, but I checked and found they are basically the same. See here: http://code.nabla.net/doc/PIL/api/PIL/PpmImagePlugin/PIL.PpmImagePlugin.PpmImageFile.html#PIL.PpmImagePlugin.PpmImageFile (OK this is an old documentation, but it shows the inheritance graph).

And if you try `image.copy()`, then the output is just `PIL.Image`. 

image

I don't think this is the optimal solution right now -- but we might keep it just for the time being..

mmda/types/document.py
```diff
@@ -107,20 +107,24 @@ def to_json(self, fields: Optional[List[str]] = None, with_images=False) -> Dict
...
+
+        # add to doc dict
+        for field in fields:
+            doc_dict[field] = [doc_span_group.to_json() for doc_span_group in getattr(self, field)]
+
+        return doc_dict
 
     @classmethod
     def from_json(cls, doc_dict: Dict) -> "Document":
-        fields = doc_dict.keys()        # TODO[kylel]: this modifies the referenced dict, not copy
```
lolipopshock, line 120: Do we want to add `doc_dict = doc_dict.copy()` to avoid changing the input?

mmda/types/names.py
```diff
@@ -6,5 +6,13 @@
 
 """
 
+# document field names
 Symbols = 'symbols'
 Images = 'images'
+
+Pages = 'pages'
```
lolipopshock, line 13: I have a question on this. We only create the fields for all the default fields that the document parser can generate, right? But we also allow users create their own fields by the annotate function, yet they can't change the fields here in their code. Would it be a bit weird? 

And FYI, we could also try something like this: https://github.com/huggingface/transformers/blob/24cbf6bc5a0b6a9bb5afdda6bb1a329ac980fa4b/src/transformers/file_utils.py#L1941

mmda/types/annotation.py
```diff
@@ -29,8 +29,8 @@ class Annotation:
     def to_json(self) -> Dict:
         pass
 
-    @abstractmethod
     @classmethod
+    @abstractmethod
```
lolipopshock, line 33: Ahh my bad 😅 thanks for fixing this!

mmda/types/box.py
```diff
@@ -18,11 +18,23 @@ class Box:
...
     @classmethod
-    def from_json(cls, box_dict) -> "Box":
-        return Box(l=box_dict['l'], t=box_dict['t'], w=box_dict['w'], h=box_dict['h'], page=box_dict['page'])
+    def from_json(cls, box_coords: List) -> "Box":
+        l, t, w, h, page = box_coords
+        return Box(l=l, t=t, w=w, h=h, page=page)
+        # return Box(l=box_coords['l'], t=box_coords['t'], w=box_coords['w'], h=box_coords['h'], page=box_coords['page'])
 
+    @classmethod
+    def small_boxes_to_big_box(cls, boxes: List['Box']) -> 'Box':
```
lolipopshock, line 32: I think we need to have a specific documentation page that specifies several different ways for "combining" boxes. 
I would call it "union" somehow..

mmda/types/annotation.py
```diff
@@ -125,7 +123,8 @@ def find(self, query: Annotation) -> List[Annotation]:
 @dataclass
 class SpanGroupIndexer(Indexer):
 
-    _index: IntervalTree = IntervalTree()
+    # careful; if write it as _index = IntervalTree(), all SpanGroupIndexers will share the same _index object
```
lolipopshock, line 126: Oh yes that's true - and this is almost the same as the list problem. Sorry for missing it earlier!

mmda/parsers/symbol_scraper_parser.py
```diff
@@ -17,51 +17,50 @@
...
-            if outfname is None:
-                outfname = os.path.join(outdir, os.path.basename(infile).replace('.pdf', '.json'))
+        if load_images:
+            raise NotImplementedError(f'Load images?')
 
-            outfile = os.path.join(outdir, outfname)
-            with open(outfile, 'w') as f_out:
+        if output_json_path:
+            with open(output_json_path, 'w') as f_out:
                 json.dump(doc.to_json(), f_out, indent=4)
```
lolipopshock, line 47: Removing `indent=4` can further reduce the storage redundancy -- maybe we can add a todo here?

mmda/parsers/symbol_scraper_parser.py
```diff
@@ -119,8 +118,8 @@ def _parse_row_head_tag(self, row_tag: str) -> Dict:
         match = re.match(pattern=r'<Line id=\"([0-9]+)\" BBOX=\"(.+)\">', string=row_tag)
         return {'id': int(match.group(1)), 'bbox': match.group(2)}
 
-    def _parse_word_head_tag(self, word_tag: str) -> Dict:
-        match = re.match(pattern=r'<Word id=\"([0-9]+)\">', string=word_tag)
+    def _parse_token_head_tag(self, token_tag: str) -> Dict:
```
lolipopshock, line 121: Just to confirm: as we are using "`token`" here, this means sscraper generated `word` might not be actual words -- probably they are split by lines or something else, right?

tests/test_parsers/test_symbol_scraper_parser.py
```diff
@@ -85,32 +85,38 @@ def setUpClass(cls):
...
-                      {'start': 25, 'end': 31, 'id': 3, 'bbox': [0.365513545751634, 0.2131515151515152, 0.04706188398692812, 0.011132455808080727]},
-                      {'start': 32, 'end': 36, 'id': 4, 'bbox': [0.4171823202614379, 0.2131515151515152, 0.027787840359477123, 0.008692116161616165]},
-                      {'start': 37, 'end': 41, 'id': 5, 'bbox': [0.44990264705882355, 0.2131515151515152, 0.03123896781045743, 0.008503431060606054]}],
-            'row': [{'start': 0, 'end': 24, 'id': 0, 'bbox': [0.24914161764705883, 0.14273611111111115, 0.302788522875817, 0.018932647727272683]},
-                    {'start': 25, 'end': 41, 'id': 1, 'bbox': [0.365513545751634, 0.2131515151515152, 0.115628069117647, 0.011132455808080727]}],
-            'sent': [],
-            'block': []
+            Symbols: 'The Smoothed Possibility\nCurves Flow RPI,\n',
+            Pages: [
+                {'_type': 'SpanGroup',
```
lolipopshock, line 104: BTW, do you like we store the `_type` in the JSON?

lolipopshock:
I left some comments to answered your questions or just to confirm something -- otherwise it looks great!

mmda/types/names.py
kyleclo: not that weird.  the purpose of this is simply to keep the library clean and without string name typos.

if someone wants to add a model to the library that populates a particular field, they can include in their PR the name of that field in this file.  

if they are adding fields in their own python sessions, then it's not our job to manage those names

mmda/types/box.py
kyleclo: i previously had it called `union` but it wasn't descriptive enough

but yes, we'll have different combining strategies

mmda/parsers/symbol_scraper_parser.py
kyleclo: added

mmda/parsers/symbol_scraper_parser.py
kyleclo: that's right.  sscraper calls it `words` but we've decided to call these `tokens`

tests/test_parsers/test_symbol_scraper_parser.py
kyleclo: not really; i think it takes up a lot of space for not much benefit -- it's very obvious whether it's a SPanGroup vs BoxGroup.  SpanGroup contains `spans` key.  i'll remove

mmda/types/document.py
kyleclo: i changed `self.fields` to just contain added fields.  we shouldnt consider `Symbols` or `Images` fields; they're more just core data representation

kyleclo: this also makes the `to_json` `from_json` methods to be easier wrt logic -- we have explicit calls to Symbols Images wrt doc_dict manipulation; and loop through all additional fields otherwise

PR #11, opened 2021-08-09
Shannons/initial model api

lolipopshock:
The organization: 

1. For each model, we'll have a `XXXDocumentPredictor` inherited from `BaseDocumentPredictor` and `XXXPredictorConfig` inherited from `BaseDocumentPredictorConfig`. 
2. The core design for `BaseDocumentPredictor` is that 
    1. the entry point `predict` takes in a `document:Document`  object and generate a list of `Annotation` 
    2. the predictor packs all the preprocessor and postprocessors via `initialize_preprocessor`, `preprocess`, `initialize_postprocessor`, and `postprocess`. 
3. I included two reference implementation (not final, just illustrative examples about how the models will be used):
    1. efficientdet model -- this is an (1) image-based and (2) wrapper model. It shows how we can wrap around existing tools and adapt them to `mmda`. 
    2. VILA models -- this is a (1) text-based and (2) "native" model (which includes all the modeling code in the `mmda` library, rather importing from somewhere else). All the modeling is incorporated in `mmda.models.vila.modeling` folder.


mmda/models/base_document_predictor.py
```diff
@@ -0,0 +1,85 @@
...
+    ###################################################################
+
+    # TODO[Shannon] Add the check for required backends in the future.
+    # So different models might require different backends:
+    # For example, LayoutLM only needs transformers, but LayoutLMv2
+    # needs transformers and Detectron2. It is the model creators'
+    # responsibility to check the required backends.
+    @property
+    @abstractmethod
+    def REQUIRED_BACKENDS(self):
```
kyleclo, line 22: this doesnt matter.  each model will have its own service defined w/ its own Dockerfile/requirements/etc.

mmda/models/base_document_predictor.py
```diff
@@ -0,0 +1,85 @@
...
+        return None
+
+    @property
+    @abstractmethod
+    def REQUIRED_DOCUMENT_FIELDS(self):
```
kyleclo, line 27: i'd remove this.  it's insufficient to make things tied to fieldnames that are agnostic to the model that produced that field.  for example, Model A produces `sents` might be different from Model B's `sents`.  if your Model C depends on Model A's `sents` and not Model B's `sents`, then just storing `sents` isnt sufficient.

for now, let's not worry about enforcing them in the implementation.  start w/ good documentation, release practices

mmda/models/base_document_predictor.py
lolipopshock: yes -- but they need to import something. And I'll try to use HF's [`LazyModule`](https://github.com/huggingface/transformers/blob/76cadb7943c8492ec481f4f3925e9e8793a32c9d/src/transformers/file_utils.py#L1953) to circumvent that issue

mmda/models/base_document_predictor.py
lolipopshock: Yup -- but that might cause some bugs? Because we need to call `doc.sents` and it's probably reasonable to check that beforehand?

mmda/models/base_document_predictor.py
kyleclo: why is this necessary?  wouldnt each model py-file have its imports defined at the top?

kyleclo (approved):
lgtm; reviewed in person

PR #13, opened 2021-08-11
bugfix; add sort() to indexer find() method; add _lt_ method to spang…

kyleclo:
bugfix for indexer
1. add sort() to indexer find() method because interval tree returns a set, not list
2. add _lt_ method to spangroups and span
3. move indexer to its own file

also some nice quality fixes
1. document's symbols are hidden method now
2. added a symbols property to SpanGroup to make easier indexing into document for strings


lolipopshock:
👍 All great

PR #14, opened 2021-08-20
symbol-scraper-parser in docker

yoganandc:


services/parsers/symbolscraper/model/prediction.py
```diff
@@ -0,0 +1,40 @@
+from typing import List
+
+from pydantic import BaseModel, Field, conint, conlist, validator
+
+
+_SPAN_BOX_LEN = 5
+
+
+class Span(BaseModel):
```
kyleclo, line 9: should `Span` and `SpanGroup` be imported from the library instead of defined here?

services/parsers/symbolscraper/setup.py
```diff
@@ -0,0 +1,30 @@
...
+    "uvicorn[standard]",
+    "pydantic",
+    "fastapi",
+    "click",
+    "python-json-logger",
+]
+
+# Add your python dependencies
+model_requirements = [
+    "mmda@git+https://github.com/allenai/mmda.git@yogic/s2agemaker-symbol-scraper"
```
yoganandc, line 16: will update this with the tag i make after merging this PR

cmwilhelm (approved):
Beautiful.

PR #18, opened 2021-08-23
example service

kyleclo:


lolipopshock:
## Explanation of this design

We hope mmda to become not only a model deployment toolkit optimized for S2, but also an open-source toolkit that many people outside S2 can have access to - this requires we package all the key functionalities into the main mmda library and make it pip-downloadable. 

However, mmda requires multiple backends/parsers/model-specific installations - it not properly handled, it would be a disaster for both the end users and deployment. So here is what we plan to handle this: 
- in mmda's `setup.py`, we design two levels of mmda dependency installation:
    - when running `pip install mmda`, it only installs the packages specified in [`install_requires`](https://github.com/allenai/mmda/blob/a656ae52b2325f8519628a436c0a27b1752ad945/setup.py#L8). And we'll make it very minimum. In this case, you might not be able to run certain parsers or models. 
    - when running `pip install mmda[xxx]` (`xxx` can be a model/parse name), it will install both the basic packages and the model specific dependencies specified in `extra_requires`. For example, when you run `pip install mmda[symbol-scraper]`, it will only install the dependencies related to `symbol-scraper`, and won't install anything else required to run other models. 
- and we will adopt the [`_LazyModule`](https://github.com/huggingface/transformers/blob/2772d3e79d66925cf4adeaffd8be610f0ab177b6/src/transformers/file_utils.py#L1957) to avoid `ImportError` when not installing deps for unused models/parsers. 

And how this design will help deployment or install services? 
-  for all the services (for example, we have `services/parsers/symbol_scraper_parser/` right now), we have two options: 
    - in the Dockerfile, we can just run `pip install mmda[xxx]` (e.g., `pip install mmda[symbol-scraper]`) to install the parser/model specific deps. 
    - when we want to package multiple steps into one single service (for example, running `symbol-scraper` as well as some sentence splitting), we can have an additional `requirements.txt` file that lists all the necessary deps, and in the Dockerfile we can run `pip install mmda && pip install -r requirements.txt`. 

How about building customized and more complicated models? Do we want to have all the models inside `mmda`? 
- No -- our key objective is to implement the most basic models inside mmda, such that it be used open-the-box. 
- And more advanced models can be implemented inside individual services folders (for example, we can do some fancy stuffs in `services/parsers/symbol_scraper_parser/` to improve the basic parser. And the dependencies will be handled within each of the service folders.

kyleclo:
discussed in person - adhering to this using a different PR for now

PR #31, opened 2021-09-08
Dockerized layoutparser

rodneykinney:
https://github.com/allenai/scholar/issues/28666

Docker container and integration tests for `mmda/predictors/lp_predictors.py`

 - Basic template copied from `s2agemaker-template` into `services/predictors/layoutparser`
 - Moved Docker container context to top level, so we can copy in the `mmda` code from the local source tree
 - No shared code yet between projects within`services`
 - The layoutparser predictor produces `BoxGroup`s which are an `Annotation`, but the `Document` class doesn't yet have a way to add them; it only manages `SpanGroup`s, so the API response cannot be deserialized into a `Document`
 - Images are passed to the API as an array of base64-encoded images, one per page.

@kyleclo @lolipopshock are we intending for `Document` to keep track of `Box` annotations in addition to span annotations?


mmda/predictors/lp_predictors.py
```diff
@@ -18,7 +19,7 @@ def __init__(self, model: lp.Detectron2LayoutModel):
     @classmethod
     def from_pretrained(
         cls,
-        config_path: str,
+        config_path: str = 'lp://PrimaLayout/mask_rcnn_R_50_FPN_3x/config',
```
rodneykinney, line 22: @lolipopshock What set of pre-trained weights do you recommend we use for production SPP?

services/predictors/layoutparser/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN pip install -r requirements.txt
+RUN pip install --upgrade .[dev]
+RUN pip install "git+https://github.com/facebookresearch/detectron2.git@v0.4#egg=detectron2"
+
+COPY services/predictors/layoutparser .
+COPY mmda mmda
+ENTRYPOINT ["/bin/bash"]
+
+
+# server image for running the API
```
rodneykinney, line 20: Using multi-stage build so that we can have a lighter-weight container (with no model downloaded) for format-checks, etc.

services/predictors/layoutparser/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM dev AS server
+
+ENV ARTIFACTS_DIR /opt/ml/model
+ENV BATCH_SIZE 1
+ENV MODEL_SERVER_TIMEOUT 60
+ENV NUM_WORKERS 1
+ENV PYTHONUNBUFFERED 1
+
+# Download the model weights
+RUN python3 entrypoint.py init
```
rodneykinney, line 30: This downloads the model weights and bakes them into the image. It adds 300MB to the image, but this image is already 5.5G, so I think it's worth it to reduce start-up time.

services/predictors/layoutparser/Makefile
```diff
@@ -0,0 +1,51 @@
+PROJECT_NAME=$(shell cat PROJECT_NAME.txt)
+
+ifdef TAG
+	OPTIONAL_TAG = -$(TAG)
+endif
+
+DEV_IMAGE_NAME=$(PROJECT_NAME)-dev$(OPTIONAL_TAG)
+SERVER_IMAGE_NAME=$(PROJECT_NAME)-server$(OPTIONAL_TAG)
```
rodneykinney, line 8: Add both image builds to the `Makefile`

services/predictors/layoutparser/integration_tests/test_invocations.py
```diff
@@ -0,0 +1,48 @@
...
+            dict(output_field="asdf:4.2"),
+            dict(output_field="fdsa:6.4"),
+            dict(output_field="qwer:-2.0"),
+        ]
+    )
+
+
+class TestInvocations(TestCase):
+    def test__invoke(self):
+        page_images = [
```
rodneykinney, line 20: End-to-end integration test. This runs in reasonable time on a CPU. (Less time than it takes to build the image, anyway)

services/predictors/layoutparser/model/instance.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel, Field
+from typing import List
+
+
+class Instance(BaseModel):
```
rodneykinney, line 5: Input model

services/predictors/layoutparser/model/prediction.py
```diff
@@ -0,0 +1,21 @@
...
+    height: float
+    page: int
+
+
+class BoxGroup(BaseModel):
+    boxes: List[Box]
+    type: str
+
+
+class Prediction(BaseModel):
```
rodneykinney, line 18: Output model

services/predictors/layoutparser/model/predictor.py
```diff
@@ -0,0 +1,68 @@
...
+    def predict_batch(self, instances: List[Instance]) -> List[Prediction]:
+        predictions = []
+
+        for instance in instances:
+            doc = Document(
+                symbols="", images=[Image.frombase64(i) for i in instance.page_images]
+            )
+            annotations = self._lp_predictor.predict(doc)
+
+            predictions.append(
```
rodneykinney, line 47: Somewhat painful translation from `mmda` model classes to `pydantic` model classes. Maybe this can be streamlined by converting to json first?

services/predictors/layoutparser/mypy.ini
```diff
@@ -0,0 +1,11 @@
+[mypy]
+check_untyped_defs = True
+no_implicit_optional = True
+plugins = pydantic.mypy
+exclude = mmda/
```
rodneykinney, line 5: Ignore errors from the base `mmda` code, since I don't want to enforce them at the `services` level. There ought to be checks at the top level, after which I can remove this.

services/predictors/layoutparser/tests/server/test_api.py
```diff
@@ -0,0 +1,206 @@
...
+TEST_PREDICTIONS = [
+    {"output": ["some\nstr", "some\nstrsome\nstr", "some\nstrsome\nstrsome\nstr"]},
+    {"output": ["someotherstrsomeotherstr", "someotherstr"]},
+]
+
+
+@patch("server.api.Instance", FakeInstance)
+@patch("server.api.Prediction", FakePrediction)
+@patch("server.api.get_predictor", get_mock_predictor)
+class TestApi(unittest.TestCase):
```
rodneykinney, line 51: Do these tests do anything useful?

services/predictors/layoutparser/verify.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "=====RUNNING INTEGRATION TESTS====="
+cleanup() {
+  trap - 0
+  set +e
+  make kill-server
+}
+trap cleanup 0
+
+make background-serve
```
rodneykinney, line 25: Similar to the `symbolscraper`'s `verify.sh`, except moved a couple of things into the Makefile.

services/predictors/layoutparser/model/predictor.py
yoganandc: does something like this https://github.com/allenai/mmda/blob/4d11cb41a8c270119bc06a5fe0e5f9ebb0da03cb/services/parsers/symbolscraper/model/predictor.py#L52
not exist for these set of mmda and pydantic models?

if not, we could add some classmethods to the pydantic classes that take in mmda models and produce pydantic variants. 

Im also okay with this being as is, even though it is painful as you say.

services/predictors/layoutparser/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+dev-image:
+	docker build -f Dockerfile --target dev -t $(DEV_IMAGE_NAME) ../../..
```
yoganandc, line 11: TIL --target

services/predictors/layoutparser/tests/server/test_api.py
yoganandc: these tests test the template code, i guess its useful to keep them in case of any regressions

yoganandc (commented):
this looks great, i just have one concern, i think in pursuit of producing reproducible images, we should change the context for the docker build back to `services/predictors/layoutparser` and have mmda be pip installed from a git tag similar to
https://github.com/allenai/mmda/blob/4d11cb41a8c270119bc06a5fe0e5f9ebb0da03cb/services/parsers/symbolscraper/setup.py#L15

if code that you're invoking has been committed since the last release, feel free to make a new one.

mmda/predictors/lp_predictors.py
lolipopshock: Thanks! I am updating layoutparser to include the some newer options here that are optimized for production. Will share with you later today.

services/predictors/layoutparser/model/predictor.py
rodneykinney: We can do this once we modify `lp_predictor` to return a `Document` class with the `BoxGroup`s added, which requires some work on the `Document` class.

rodneykinney:
Per discussion, decided to let the docker build pull in the mmda code from the current commit.

PR #36, opened 2021-09-14
Heuristic predictor for detecting hyphenated words

rauthur:
Example usage:

```python
SSCRAPER = "/workspaces/mmda/SymbolScraper/bin/sscraper"

from mmda.parsers.symbol_scraper_parser import SymbolScraperParser
from mmda.predictors.heuristic_predictors.dictionary_word_predictor import (
    DictionaryWordPredictor,
)
from mmda.types.document import Document

INPUT_PDF_PATH = "/workspaces/mmda/docs/maml.pdf"
OUTPUT_JSON_PATH = "/workspaces/mmda/output/maml.json"
TEMP_PATH = "/workspaces/mmda/tmp"

ssparser = SymbolScraperParser(sscraper_bin_path=SSCRAPER)
doc: Document = ssparser.parse(
    input_pdf_path=INPUT_PDF_PATH, output_json_path=OUTPUT_JSON_PATH, tempdir=TEMP_PATH
)

words = DictionaryWordPredictor("/workspaces/mmda/resources/english.txt")
result = words.predict(doc)

print([w.text for w in result])
```

Demonstration, for the following abstract (taken from http://proceedings.mlr.press/v70/finn17a/finn17a.pdf)

maml-text

Yield results include:

```text
// grabbing terms from abstract only
'We', 'propose', 'an', 'algorithm', 'for', 'meta-learning', 'that', 'is', 'model-agnostic,', 
'in', 'the', 'sense', 'that', 'it', 'is', 'compatible', 'with', 'any', 'model', 'trained', 'with', 
'gradient', 'descent', 'and', 'applicable', 'to', 'a', 'variety', 'of', 'different', 'learning', 
'problems,', 'including', 'classiﬁcation,', 'regression,', 'and', 'reinforcement', 'learning.', 
'The', 'goal', 'of', 'meta-learning', 'is', 'to', 'train', 'a', 'model', 'on', 'a', 'variety', 'of',
 'learning', 'tasks,', 'such', 'that', 'it', 'can', 'solve', 'new', 'learning', 'tasks', 'using', 
'only', 'a', 'small', 'number', 'of', 'training', 'samples.', 'In', 'our', 'approach,', 'the', 
'parameters', 'of', 'the', 'model', 'are', 'explicitly', 'trained', 'such', 'that', 'a', 'small',
 'number', 'of', 'gradient', 'steps', 'with', 'a', 'small', 'amount', 'of', 'training', 'data', 
'from', 'a', 'new', 'task', 'will', 'produce', 'good', 'generalization', 'performance', 'on',
 'that', 'task.', 'In', 'effect,', 'our', 'method', 'trains', 'the', 'model', 'to', 'be', 'easy', 'to',
 'ﬁne-tune.', 'We', 'demonstrate', 'that', 'this', 'approach', 'leads', 'to', 'state-of-the-art', 
'performance', 'on', 'two', 'few-shot', 'image', 'classiﬁcation', 'benchmarks,', 'produces',
 'good', 'results', 'on', 'few-shot', 'regression,', 'and', 'accelerates', 'ﬁne-tuning', 'for', 
'policy', 'gradient', 'reinforcement', 'learning', 'with', 'neural', 'network', 'policies.'
// continues...
```


mmda/predictors/heuristic_predictors/dictionary_word_predictor.py
```diff
@@ -0,0 +1,180 @@
...
+
+    _dictionary: Optional[set[str]] = None
+
+    def __init__(self, dictionary_file_path: str) -> None:
+        """Build a predictor that indexes the given dictionary file.
+        A dictionary is simply a case-sensitive list of words as a text file.
+        Words should be lower-case in the dictionary unless they are invalid
+        as a lower-case word (e.g., a person's name).
+
+        For an example, see https://github.com/dwyl/english-words (words_alpha.txt)
```
lolipopshock, line 23: I think we could cache it to s2 aws and if the dictionary_file_path is not set, we can automatically have the default version downloaded.

rauthur: I think that giving a reference here is the right thing. This way the consume can decide if the license of the consumed dictionary is appropriate. Currently the example repository is Unlicense which would be fine in general, but there is an underlying reference to the original copyright holder being someone else. So, I think users should be able to decide what dictionary reference to use.

mmda/predictors/heuristic_predictors/dictionary_word_predictor.py
```diff
@@ -0,0 +1,180 @@
...
+    def dictionary(self) -> set[str]:
+        """Global dictionary and not document specific. This dictionary is the basis for
+        finding words and will be appended with document-level tokens.
+
+        Returns:
+            set[str]: A set of words in the dictionary. For proper names, the dictionary
+            may contain just the title-cased word (e.g., "Russell" and not "russell").
+        """
+        if not self._dictionary:
+            with open(self.dictionary_file_path, "r") as source:
```
kyleclo, line 47: maybe incluede a TODO comment about format checks

tests/test_predictors/test_dictionary_word_predictor.py
```diff
@@ -0,0 +1,189 @@
+"""
+
+Tests for DictionaryWordPredictor
+
+@rauthur
+
+
+"""
```
kyleclo, line 8: can u add a test fixture of an example dictionary file (mayb like 5 lines)

kyleclo (approved):
just 2 comments

PR #38, opened 2021-09-16
[fix] Return relative coordinates in LP predictor

lolipopshock:
Previously, LP predictor returns the absolute coordinates, which is incompatible with the rest of the library. In this update, 
1. LP predictor returns the relative coordinates
2. Fix the coordinate conversion for VILA models 
3. Update the VILA example to incorporate main mmda updates


kyleclo (approved):
ok

PR #42, opened 2021-09-23
Kylel/2021 09 17/cleanup

kyleclo:
some code cleanup

shouldnt affect much but main things are that certain fields are now double `__` in anticipation for something @lolipopshock and i discussed about making visual BoxGroups indexable via single `._` and textual SpanGroups indexable via your usual `.`

for example `Document.blocks` would be SpanGroups and `Document._blocks` would be the corresponding box groups


rodneykinney (approved):

requirements.txt
```diff
@@ -1 +1,2 @@
+tqdm
```
yoganandc, line 1: add this to setup.py pls, not here

yoganandc (approved):

mmda/types/document.py
```diff
@@ -25,65 +25,71 @@
...
+        # 2) register fields into Document
         for field_name, annotations in kwargs.items():
-            if len(annotations) == 0: 
+            if len(annotations) == 0:
                 warnings.warn(f"The annotations is empty for the field {field_name}")
-                continue 
+                continue
+
+            # TODO: why is this good idea?
             annotations = deepcopy(annotations)
```
lolipopshock, line 69: We don't want to overwrite the original annotations, I think?

lolipopshock (approved):

PR #48, opened 2021-09-29
VILA predictor service

rodneykinney:
https://github.com/allenai/scholar/issues/29184

REST service for VILA predictors

Any code that I didn't comment on in this PR is s2age-maker boilerplate.


lolipopshock:
Hey @rodneykinney ! Thank you very much for this PR! 

PS: I've just uploaded the VILA repo and added some functional improvements -- could you also try to install VILA directly from the GitHub repo's master branch (`pip install git+https://github.com/allenai/VILA@master#egg=vila`) and see if there is any bug in your service? If not, I can then go ahead and release the new version of VILA and you can work with the new version in the future. Thanks a ton!!

rodneykinney:
@lolipopshock My tests passed when installing vila from git@master, so I think you are okay to release the new version.

mmda/predictors/hf_predictors/utils.py
```diff
@@ -15,15 +15,16 @@ def shift_index_sequence_to_zero_start(sequence):
 
 
 def get_visual_group_id(token: SpanGroup, field_name: str, defaults=-1) -> int:
-
+    if not hasattr(token, field_name):
```
rodneykinney, line 18: Check for existence of field.

mmda/predictors/hf_predictors/utils.py
```diff
@@ -15,15 +15,16 @@ def shift_index_sequence_to_zero_start(sequence):
...
+        return defaults
     field_value = getattr(token, field_name)
     if len(field_value) == 0 or field_value[0].id is None:
         return defaults
     return field_value[0].id
 
 
 def convert_document_page_to_pdf_dict(
-    document: Document, page_width: int, page_height: int
+    document: Document
```
rodneykinney, line 27: I got rid of the bbox-coordinate scaling in this method. For one thing, the coordinates were absolute to begin with in my case. For another thing, I was using LayoutLM model that doesn't use the bounding boxes anyway, so loading the page dimensions is a waste.

If confusion between relative and absolute coordinates becomes a problem, I recommend we add a flag to the `Box` class to indicate which coordinates are used in that instance.

mmda/predictors/hf_predictors/vila_predictor.py
```diff
@@ -147,11 +147,7 @@ def predict(self, document: Document) -> List[Annotation]:
         page_prediction_results = []
         for page_id, page in enumerate(tqdm(document.pages)):
 
-            page_width, page_height = document.images[page_id].size
-
-            pdf_dict = convert_document_page_to_pdf_dict(
-                page, page_width=page_width, page_height=page_height
-            ) 
+            pdf_dict = convert_document_page_to_pdf_dict(page)
```
rodneykinney, line 150: Here's where it doesn't seem right to require loading the images just to re-scale the coordinates.

Another way to get the page sizes is from the `Document.pages` I wasn't clear if the page dimensions that come from symbolscraper are the entire page or are just the bounding box that contains all the tokens (page size minus the margins). Having the complete page would be prefereable, IMO.

services/predictors/vila/Makefile
```diff
@@ -0,0 +1,57 @@
...
+format-check: build-image
+	docker run --rm --entrypoint /bin/bash $(IMAGE_NAME) -c 'black --exclude "mmda/.*" --check .'
+
+reformat: build-image
+	docker run -v $(shell pwd):/opt/ml/code --rm --entrypoint /bin/bash $(IMAGE_NAME) -c 'black --exclude "mmda/.*" .'
+
+format: build-image
+	docker run --rm -v $(shell pwd):/opt/ml/code --entrypoint bash $(IMAGE_NAME) -c 'black .'
+
+model-weights:
```
rodneykinney, line 43: Model weights are downloaded from S3 by the Makefile and baked into the Docker image.

services/predictors/vila/server/api_model.py
```diff
@@ -0,0 +1,42 @@
+"""
+Pydantic equivalents of classes from mmda.types
```
rodneykinney, line 2: These serialize into JSON the same as the `mmda` types, so are only used in the outmost layer of the API request handler.

Gave up trying to inject classes into the `mmda` library that would be usable by this API.

services/predictors/vila/server/api_model.py
```diff
@@ -0,0 +1,42 @@
...
+        assert len(v) == 5, "Box must have format [x, y, w, h, page]"
+        assert int(v[4]) == v[4], "Page must be an integer value"
+
+
+class SpanGroup(BaseModel):
+    spans: List[Span]
+    type: Optional[str]
+
+    @classmethod
+    def from_mmda(cls, grp: MmdaSpanGroup):
```
rodneykinney, line 31: Convert back and forth between `mmda` types.

rodneykinney:
I didn't experiment with different model weights, but here's what I found applying the grotoap weights listed in the Makefile, against the `services/parsers/symbolscraper/integration_tests/data/arxiv-1906.08632.pdf` file.

- The model produces just two labels: `type=2` and `type=3`. I couldn't determine what they were supposed to mean.
- symbolscraper produces spurious `fraction(-)` tokens at the end of some lines (including the middle of the title on the header page). I suppose the model would learn to ignore those tokens if trained on symbolscraper output, so this is a hint that models trained on pdfplumber will have problems.

services/predictors/vila/server/api_model.py
kyleclo: hmm, a bit unfortunate that this is necessary but I guess fine for now. open to ideas about how to consolidate.

mmda/predictors/hf_predictors/vila_predictor.py
kyleclo: good point.  I think getting it from `.images` at least guarantees that the bbox dimensions are for the whole page.  with deriving them from `.pages` which are of type `SpanGroup`, it's unclear what this bbox should be wrt margins.  i think this will always be ambiguous for `.pages` but at least from `.images` it should be very clear.

hence my preference to go from `.images`

mmda/predictors/hf_predictors/utils.py
kyleclo: can explain why not raise Exception?

kyleclo (changes requested):
some questions

lolipopshock:
I have some comments, and I will add later today. Thanks!!

rodneykinney:
To get the dimensions of a page, how about making a BoxGroup annotation with the full size of each page? This could be expected output from a parser. Even if the parser figures out the dimensions by inspecting the page images, it's better than passing around full base64-encoded image data.

rodneykinney:
In `visual_group_id`, there's already code to populate with a default value if a field is missing, so I was following that pattern. In the case of the LayoutLM model, blocks weren't needed so I didn't want to throw an exception if I didn't supply them. 

If blocks are required, then it would be better to throw than to populate with default values, but some VILA models require blocks and some don't. Maybe this is an argument for using a different predictor class to represent LayoutLM models. Otherwise, to solve this, we would need a way to declare what a model expects to find on an input Document at a model-instance level.

mmda/predictors/hf_predictors/utils.py
lolipopshock: I don't think we should remove this for two reasons: 

1. the image sizes will not only be used for correcting the block sizes, but also modifying "correcting" token sizes. For original LayoutLM models, it requires to scale all the `token` bounding box to [0, 1000] (which we haven't implemented yet (should be in line 56 below). (for our model, we just use the original coordinates for convenience). 
2. more broadly, this should be a model-specific parameter, rather than be stored in the document object.

mmda/predictors/hf_predictors/vila_predictor.py
lolipopshock: I have the same opinion as Kyle!

services/predictors/vila/server/predictor.py
```diff
@@ -0,0 +1,46 @@
...
+from mmda.predictors.hf_predictors.vila_predictor import SimpleVILAPredictor
+from mmda.types.document import Document
+from mmda.types.annotation import SpanGroup
+
+
+class PredictorConfig(BaseSettings):
+    """Uninitialized fields must be set via environment variables"""
+
+    ARTIFACTS_DIR: str
+    label_names = {2: "label2", 3: "label3"}
```
lolipopshock, line 15: So in the new VILA models, I've managed to store both the label_map and vila related configurations inside the model config file, e.g.,
- https://huggingface.co/allenai/ivila-block-layoutlm-finetuned-docbank/blob/main/config.json#L11
- https://huggingface.co/allenai/ivila-block-layoutlm-finetuned-docbank/blob/main/config.json#L55

And the code is to set automatically load the configurations, see:
- https://github.com/allenai/VILA/blob/e53cf1ef1edc8623162603b373e168f3d2aa1758/src/vila/predictors.py#L59
- https://github.com/allenai/VILA/blob/e53cf1ef1edc8623162603b373e168f3d2aa1758/src/vila/dataset/preprocessors/config.py#L11

These updates haven't been incorporated into the current mmda VILA predictors -- but should be easily implemented in the new few updates.

lolipopshock:
> To get the dimensions of a page, how about making a BoxGroup annotation with the full size of each page? This could be expected output from a parser. Even if the parser figures out the dimensions by inspecting the page images, it's better than passing around full base64-encoded image data.

I see the issue why you'd like to directly access to the figure sizes. how about this: 
- we can have a `.sizes`  or `._sizes` for each doc object. 
- it can be explicitly set, or if not, it can load the data from the images. 

CC @kyleclo

lolipopshock:
> In `visual_group_id`, there's already code to populate with a default value if a field is missing, so I was following that pattern. In the case of the LayoutLM model, blocks weren't needed so I didn't want to throw an exception if I didn't supply them.
> 
> If blocks are required, then it would be better to throw than to populate with default values, but some VILA models require blocks and some don't. Maybe this is an argument for using a different predictor class to represent LayoutLM models. Otherwise, to solve this, we would need a way to declare what a model expects to find on an input Document at a model-instance level.

Agree -- and I think a better solution is to have another predictor for LayoutLM models ...

rodneykinney:
I’m now getting an error ` AttributeError: module 'layoutparser' has no attribute 'AutoLayoutModel'` at [this line](https://github.com/allenai/mmda/blob/main/mmda/predictors/lp_predictors.py#L36) @lolipopshock was there an update to the layoutparser library?

lolipopshock:
> I’m now getting an error ` AttributeError: module 'layoutparser' has no attribute 'AutoLayoutModel'` at [this line](https://github.com/allenai/mmda/blob/main/mmda/predictors/lp_predictors.py#L36) @lolipopshock was there an update to the layoutparser library?

Yeah please update to the latest v0.3.2 version :)

mmda/types/annotation.py
```diff
@@ -89,7 +89,7 @@ class SpanGroup(Annotation):
 
     @property
     def symbols(self) -> List[str]:
-        return [self.doc._symbols[span.start : span.end] for span in self.spans]
+        return [self.doc.symbols[span.start : span.end] for span in self.spans]
```
rodneykinney, line 92: Assume this was an oversight.

mmda/types/box.py
```diff
@@ -12,8 +12,8 @@
 
 def is_overlap_1d(start1: float, end1: float, start2: float, end2: float) -> bool:
     """Return whether two 1D intervals overlaps"""
-    assert start1 < end1
-    assert start2 < end2
+    assert start1 <= end1
```
rodneykinney, line 15: Zero-width bounding boxes cause an error at this line. Is it an error to produce these? I'm inclined to say they're okay in principle. If we don't want to allow them, they should be detected/removed at some earlier point.

setup.py
```diff
@@ -5,6 +5,6 @@
     version="0.0.2",
     python_requires=">= 3.7",
     packages=setuptools.find_packages(include=["mmda.*"]),
-    install_requires=["intervaltree", "pdf2image", "pdfplumber"],
+    install_requires=["intervaltree", "pdf2image", "pdfplumber", "pandas", "requests"],
```
rodneykinney, line 8: `pandas` needed to import `mmda.parsers.pdfplumber_parser`. `requests` needed to call local Docker containers.

rodneykinney:
Did some refactoring of the layoutparser service here too. Freeze dependencies, remove some intermediate data classes.

Changes are now limited to `services/predictors` except for the latest couple of comments. @lolipopshock @kyleclo ready for another look.

PR #49, opened 2021-09-30
Use Grobid header API method to get title/abstract

rauthur:
Basic "parser" that posts a PDF to a remote (or localhost) Grobid instance. The header parser provided here only requests header information and also provides title and abstract annotations. 

I've introduced a Parser protocol because I believe this makes more sense than a base class that already has conflicts in passed parameters. For example, Grobid doesn't care about any DPI setting.

If some parsers share base functionality then they can have a parent class that exists at a non-root level. Eventually BaseParser should be removed.

### DIFF
--- a/mmda/parsers/grobid_parser.py
+++ b/mmda/parsers/grobid_parser.py
@@ -1,0 +1,100 @@
+import io
+import xml.etree.ElementTree as et
+
+import requests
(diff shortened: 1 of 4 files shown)


kyleclo (approved):
lgtm - need to follow-up w refactor of otehr Parsers

PR #53, opened 2021-10-07
Dockerized pipeline

rodneykinney:
Pattern for wrapping services in a lighter-weight containers. 

- Replace full sub-projects in the `services` directory with a `Dockerfile` plus a `xxx_api.py` file for each service.
- Add a docker-compose file that build the services plus a python container for running a REPL or scripts
- Add `pipeline.py` sample end-to-end pipeline script using the services
- Add `run-pipeline.sh` file to run pipeline

@rauthur


docker/pipeline.yml
```diff
@@ -0,0 +1,28 @@
+version: '3.7'
+services:
+  symbolscraper:
+    container_name: mmda-sscraper
+    build:
+      context: ..
+      dockerfile: docker/parsers/symbolscraper/Dockerfile
+
+#  layoutparser:
```
rodneykinney, line 9: Haven't added these yet. If we agree on this direction, I'll complete it.

PR #59, opened 2021-10-13
[wip] Move PDF->Image functionality *out* of Parser into Rasterizer

kyleclo:
1. Rasterizer handles PDF->Image.  It's separate from Parser.
2. Remove dpi=72 default value.
3. Clean up PIL.Image import names to be less confusing
4. Update library tests
5. Update SScraper predictor service to use Rasterizer

### DIFF
--- a/mmda/parsers/parser.py
+++ b/mmda/parsers/parser.py
@@ -10,7 +10,6 @@ from abc import abstractmethod
 from typing import List, Optional, Protocol, Union
 
 from mmda.types.document import Document
-from mmda.types.image import load_pdf_images_from_path
 
 
 class Parser(Protocol):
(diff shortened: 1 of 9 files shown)


mmda/parsers/parser.py
```diff
@@ -10,62 +10,20 @@
...
 from mmda.types.document import Document
-from mmda.types.image import load_pdf_images_from_path
 
 
 class Parser(Protocol):
     @abstractmethod
-    def parse(self, input_pdf_path: str, **kwargs) -> Document:
+    def parse(self,
+              input_pdf_path: str,
+              output_json_path: Optional[str] = None,
```
rodneykinney, line 19: If `output_json_path` is just to write the serialized `Document`, that doesn't seem like the responsibility of the parser. Also not consistent with the other predictors.

A more general interface would be a byte stream instead of a file path for the PDF. For example, an API might get the data from the request body without writing it to a temp file. Then the parser implementations would create a temp file from the byte stream if they needed to.

PR #62, opened 2021-10-26
Fix VLUE evaluation for rasterizer + add different parser

rauthur:


mmda/types/document.py
```diff
@@ -68,22 +73,54 @@ def annotate(self, is_overwrite: bool = False, **kwargs: List[Annotation]) -> No
...
+                raise NotImplementedError(
+                    f"Unsupported annotation type {annotation_type} for {field_name}"
+                )
 
             # register fields
             setattr(self, field_name, span_groups)
             self.__fields.append(field_name)
 
-    def _annotate_span_group(self, span_groups: List[SpanGroup], field_name: str) -> List[SpanGroup]:
+    def annotate_images(
```
lolipopshock, line 99: I think we could combine the `annotate_images` function to the `annotate` function, right? (We can even just hard-code the behavior for `doc.annotate(images=xx)`)

If we split `annotate_images` from `annotate`, I think the API just becomes too confusing for people.

rauthur: I originally did this. However, the implementation is too messy in my opinion b/c images will need special exceptions to some of the rules (is not a field name). Further, Image is not an Annotation. Until that changes the type hierarchy doesn't make sense to include in the same method.

I would propose a larger change to document where a document is either parsed by just a parser resulting in a document that has no images OR parsed by a parser + a rasterizer in which case images are added. If we made this refactor then we could revert Document from this PR and rely on the parser to delegate to a rasterizer.

Specifically the user can select to use a token parser or a token parser + image provider.

```python
parser_only = PDFPlumberParser()
doc_without_images = parser_only.parse("some.pdf")

parser_and_rasterizer = DualParser(parser=PDFPlumberParser(), rasterizer=PDF2ImageRasterizer())
doc_with_images = parser_and_rasterizer.parse("some.pdf")
```

/cc @kyleclo

lolipopshock: I am on board with the `DualParser` idea -- actually I was thinking about something similar. 

Speaking of the API design, I think an alternative could be:


```python

# specific PDF parsers (when people care about which one to use)
pdf_parser_1 = PDFPlumberParser()
pdf_parser_2 = SymbolScraperParser()

# generalized PDF parsers
pdf_parser = PDFParser() # use a default parser
pdf_parser = PDFParser(parser='pdfplumber')# specify a parser backend 

pdf_parser = PDFParser(rasterizer=True) # use a default rasterizer
pdf_parser = PDFParser(parser='pdfplumber', rasterizer='pdf2image')# specify a parser backend + rasterizer with default args

pdf_parser = PDFParser(parser= PDFPlumberParser(some_args), rasterizer= PDF2ImageRasterizer(some_args)) # specify a parser backend + rasterizer with customized args
```

kyleclo: I disagree with `doc.annotate()` being overloaded for the reasons @rauthur described above.  

I think we should hold off on `DualParser`.  I still want the ability to augment an existing Document with images after-the-fact.  The DualParser idea doesn't solve this problem, and it's only slightly less boilerplate code to write than `doc.annotate_images()`

If we're worried about confusion between `annotate` being used for both fields and underlying representations like images, then just rename that method something extremely explicit -- `doc.attach_page_images()`.

kyleclo (approved):
lgtm

mmda/types/document.py
rauthur: @lolipopshock I like the idea of a factory for users or we could just provide something like `Parser.default` that returns a base configuration with all requirements in setup.py under a "defaults" optional install or something like that.

PR #63, opened 2021-11-02
Fix if the coordinates are larger than 1000 

lolipopshock:
[Temporary solution for now..] Show warning and fix the coordinates when necessary

### DIFF
--- a/mmda/predictors/hf_predictors/utils.py
+++ b/mmda/predictors/hf_predictors/utils.py
@@ -66,11 +66,17 @@ def convert_document_page_to_pdf_dict(
         for token in document.tokens
     ])
 
-    if bbox.max() > 1000:
+    if bbox.max() > 1000 or bbox.min() < 0:
         warnings.warn(
             "The absolute coordinates are larger than 1000. "
             "Please check whether you've set the page_width and page_height correctly."
         )
+    if bbox.min() < 0:
+        warnings.warn(
+            "The absolute coordinates are smaller than 0. "
+            "You might need to check the whether the PDF parser works as expected."
+        )
+    if bbox.max() > 1000 or bbox.min() < 0:
         bbox = bbox.clip(0, 1000)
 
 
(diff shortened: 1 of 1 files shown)


rauthur:
Can you explain why clipping is a fix? I ran into the max 1000 dim error when using a higher DPI. The "fix" in my case was to use a lower DPI. Is it even possible to use LayoutParser with larger DPI? Is that the constraint?

lolipopshock:
> Can you explain why clipping is a fix? I ran into the max 1000 dim error when using a higher DPI. The "fix" in my case was to use a lower DPI. Is it even possible to use LayoutParser with larger DPI? Is that the constraint?

Yes I agree with you -- this is a bad strategy. For now it is used for some hot bug fixes (just want to get a commit/PR created for reference)..

kyleclo:
First, since @yoganandc is the MMDA user who found this bug, we need to make sure @yoganandc approves this fix.    Without his approval, I'm hesitant to merge this.

Second, if lower DPI is a solution for this as @rauthur , we shouldn't merge this temp fix; just require lower DPI usage

rauthur:
I was experimenting with Tesseract OCR which recommends a DPI of at least 150. I tried to use the same images with layout parser, ran into this error and changed the DPI back to 72. If @yoganandc doesn't have luck at 72 then it could just be a massive PDF (in terms of dimensions). I'm not sure what's appropriate in that case.

rauthur:
@lolipopshock Do you have an outline of a fix that would be better in this case? Just want to get a rough idea of steps even if we don't change anything now.

lolipopshock:
> @lolipopshock Do you have an outline of a fix that would be better in this case? Just want to get a rough idea of steps even if we don't change anything now.

Thanks for checking this, @rauthur ! Here is my plan:
1. In the future, we won't use absolute coordinates for modeling, and we'll normalize them to some ranges, which is specified by the models.
2. So for example, for a box coordinate (12,24,30,50), page_size (200, 200), and the box range specified in the model (by default is [0, 1000]), it will convert the box coordinate to [60, 120, 150, 250], and then send to the models. This will ensure correct values regardless of the used DPIs.

lolipopshock:
Merge with the latest update from Kyle, as a reference for running eval

lolipopshock:
Fixed in #77

PR #66, opened 2021-11-16
flag to disable tqdm

yoganandc:


rodneykinney (commented):
What problem is the progress bar causing? How about disabling it [globally](https://stackoverflow.com/a/67238486) instead? To me, this is also a non-intuitive side-effect from a `debug=True` flag, which I would expect to generate more verbose logs.

yoganandc:
debug=True causes tqdm to not be disabled (and therefore the logs become more verbose). tqdm doesn't play well with datadog and im trying to ingest the sagemaker-deployed models' logs into dd.

feels much less straight-forward to hijack the constructor of a third party library to me, but we've done what you suggest before so I can roll with it.

PR #71, opened 2022-03-28
Updated example to use rasterizer APIs

soldni:
Example was a little bit out of date because it did not use `PDF2ImageRasterizer`. Because of that, `LayoutParserPredictor` couldn't find any regions.


lolipopshock (approved):
LGTM! @kyleclo  But perhaps the APIs will be changed again in the future?

lolipopshock:
OK merge it :D

PR #72, opened 2022-04-12
Additional checks for pdfplumber parser

lolipopshock:
Sometimes PDF Parsers will return token coordinates with negative or extra large values (larger than the page_width/height). This is not a bug but a feature designed in the "printing commands" of PDF grammars. However, this will break our models and many downstream processing. 

This issue aims to fix this issue via adding additional checks for PDFPlumber Parsers. It brings similar clipping functions in the VILA repo https://github.com/allenai/VILA/blob/6a5c6783b75eefaeaed01b67bc3be025add9cde3/src/vila/pdftools/pdfplumber_extractor.py#L204 to the mmda PDFPlumberParsers to increase the robustness of token parsers. 

Some examples: 

Previously a token will have a box like (the box is outside the page range)
```python
box=Box(l=0.10764222310323279, t=-0.35593721269999645, w=0.02696391589783563, h=0.009478672985782012, page=7)
```
While after this fix, it will be
```python
Box(l=0.10751146308724832, t=0.0, w=0.026931161073825507, h=0.0, page=None)
```

### DIFF
--- a/mmda/parsers/pdfplumber_parser.py
+++ b/mmda/parsers/pdfplumber_parser.py
@@ -183,14 +183,16 @@ class PDFPlumberParser(Parser):
         page_tokens = [
             {
                 "text": token["text"],
-                "bbox": Box.from_coordinates(
+                "bbox": Box.from_pdf_coordinates(
                     x1=float(token["x0"]),
                     y1=float(token["top"]),
                     x2=float(token["x1"]),
                     y2=float(token["bottom"]),
+                    page_width=page.width, 
+                    page_height=page.height,
                     page=page_index,
                 ).get_relative(
-                    page_width=float(page.width), page_height=float(page.height)
+                    page_width=page.width, page_height=page.height
                 ),
             }
             for token in token_data
--- a/mmda/types/box.py
+++ b/mmda/types/box.py
@@ -63,9 +63,9 @@ class Box:
         _x1, _x2 = np.clip([x1, x2], 0, page_width)
         _y1, _y2 = np.clip([y1, y2], 0, page_height)
 
-        if _x2 > _x1:
+        if _x2 < _x1:
             _x2 = _x1
-        if _y2 > _y1:
+        if _y2 < _y1:
             _y2 = _y1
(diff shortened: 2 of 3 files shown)


yoganandc (approved):
seems like CI is failing on main so ✅

lolipopshock:
> seems like CI is failing on main so ✅

😂 what?? anything wrong with the CI?

yoganandc:
yeah this commit broke it https://github.com/allenai/mmda/actions/runs/1974574941

but i personally dont mind if you merge without investigating what's going on since you didnt cause it

yoganandc (commented):
so i tried to verify that everything works as expected by running this script
https://github.com/allenai/mmda/blob/main/pipeline/run_local.py

against this pdf - [page0.pdf](https://github.com/allenai/mmda/files/8476870/page0.pdf)

from this branch (better-pdfplumber-parser), I get
```
Title    = ''
Abstract = ''
```

in main branch, i get 
```
Title    = 'Dynamics of stochastic gradient descent for two-layer neural networks in the teacher-student setup'
Abstract = 'Abstract Deep neural networks achieve stellar generalisation even when they have enough parameters to easily ﬁt all their training data. We study this phenomenon by analysing the dynamics and the performance of over-parameterised two-layer neural networks in the teacher-student setup, where one network, the student, is trained on data generated by another network, called the teacher. We show how the dynamics of stochastic gradient descent (SGD) is captured by a set of differential equations and prove that this description is asymptotically exact in the limit of large inputs. Using this framework, we calculate the ﬁnal generalisation error of student networks that have more parameters than their teachers. We ﬁnd that the ﬁnal generalisation error of the student increases with network size when training only the ﬁrst layer, but stays constant or even decreases with size when training both layers. We show that these different behaviours have their root in the different solutions SGD ﬁnds for different activation functions. Our results indicate that achieving good generalisation in neural networks goes beyond the properties of SGD alone and depends on the interplay of at least the algorithm, the model architecture, and the data set. Deep neural the these'
```

PR #75, opened 2022-04-15
Defines a TIMO Interface for Vila.

cmwilhelm:
For internal AI2/S2-consumption only.
Implements an interface for S2's
production model packaging toolchain.


cmwilhelm:
@yoganandc This is almost a straight copypasta from the relevant code in SPP.

@kyleclo This is what a timo-ified model looks like. Relevant parts from our discussion yesterday are sequestering this new code into a top-level module that is explicitly labeled as ai2-internal.

ai2_internal/timo/config.yaml
```diff
@@ -0,0 +1,34 @@
...
+    # These are the 'extras_require' keys in your setup.py.
+    extras_require: ["dev", "vila_predictors"]
+
+    # Full S3 path to tar.gz'ed artifacts archive, nullable
+    artifacts_s3_path: s3://ai2-timo-registry/model-artifacts/vila/v0.tar.gz
+
+    # Version of python required for model runtime, e.g. 3.7, 3.8, 3.9
+    python_version: 3.8
+
+    cuda: False
```
kyleclo, line 26: can explain why false? what does this do?

ai2_internal/timo/vila/test_fixtures/test_blocks.json
```diff
@@ -0,0 +1 @@
+{"groups": [{"boxes": [{"left": 0.22829851138046364, "top": 0.39771683047516176, "width": 0.5444175371157578, "height": 0.21800039272115687, "page": 0}], "type": "Text"}, {"boxes": [{"left": 0.17708243575750612, "top": 0.6380726592709319, "width": 0.645345301409952, "height": 0.11416552765200837, "page": 0}], "type": "Text"}, {"boxes": [{"left": 0.1833843779719733, "top": 0.7508946428395281, "width": 0.6508047066482843, "height": 0.07361510305693655, "page": 0}], "type": "Text"}, {"boxes": [{"left": 0.18469316819134882, "top": 0.8286216042258523, "width": 0.6653114268982333, "height": 0.07351931177004419, "page": 0}], "type": "Text"}, {"boxes": [{"left": 0.46725164675245096, "top": 0.3638638583096591, "width": 0.06918848573771957, "height": 0.014146631414240057, "page": 0}], "type": "Title"}, {"boxes": [{"left": 0.18218301012625102, "top": 0.11883464967361604, "width": 0.644795985003702, "height": 0.0535144516915986, "page": 0}], "type": "Title"}, {"boxes": [{"left": 0.21118448294845282, "top": 0.24982265510944404, "width": 0.5897238737617443, "height": 0.08330792128437697, "page": 0}], "type": "Text"}]}
```
kyleclo, line 1: minor, but just for ease of viewing, can u save this JSON w/ `json.dump(..., indent=4)` or something like that?

ai2_internal/timo/vila/interface.py
```diff
@@ -0,0 +1,123 @@
...
+        self._load_model()
+
+    def _load_model(self) -> None:
+        self._predictor = IVILAPredictor.from_pretrained(
+            self._artifacts_dir,
+            added_special_sepration_token="[BLK]",
+            agg_level="row",
+        )
+
+    def predict_batch(self, instances: List[Instance]) -> List[Prediction]:
```
kyleclo, line 105: do u have sense of how we shd handle difference in this predict_batch() with or without a GPU?

ai2_internal/timo/vila/integration_test.py
```diff
@@ -0,0 +1,94 @@
+"""
+Write integration tests for your model interface code here.
+
+The TestCase class below is supplied a `container`
```
kyleclo, line 4: can u include instructions somewhere for how to run these tests locally, including obtaining the container that this should point to?

ai2_internal/timo/config.yaml
cmwilhelm: If set to True it'll build a docker image that can leverage a GPU through nvidia-docker. I believe @rodneykinney found that GPU operations accounted for a minority of inference time when he was operationalizing this previously with s2agemaker (which is the same underlying set of templates). Turning this off builds a much smaller and simpler docker image.

ai2_internal/timo/vila/test_fixtures/test_blocks.json
cmwilhelm: 👍

ai2_internal/timo/vila/integration_test.py
cmwilhelm: Sure. This is documented in the `tt` user guide over here: https://github.com/allenai/timo/blob/main/docs/timo-tools/userguide.md#step-5-add-some-integration-tests

But I can make a reference to those docs in this docstring.

yoganandc (approved):

ai2_internal/timo/config.yaml
rodneykinney: What I remember is that most of the time was spent in `convert_document_page_to_pdf_dict`. I suspect that method can be made more efficient, possibly at the cost of making the input data shape less convenient. But my conclusion was that with the current code, using a GPU is not worth the cost.

PR #76, opened 2022-04-22
VILA models crashing when bounding boxes are not int

soldni:
VILA models crashing when bounding boxes are not int

Because of changes in #69 , bounding boxes are now float instead of int, which VILA does not like:

```bash
  File "/Users/lucas/miniforge3/envs/pdod/lib/python3.10/site-packages/mmda/predictors/hf_predictors/vila_predictor.py", line 161, in predict
    model_outputs = self.model(**self.model_input_collator(model_inputs))
  File "/Users/lucas/miniforge3/envs/pdod/lib/python3.10/site-packages/mmda/predictors/hf_predictors/vila_predictor.py", line 178, in model_input_collator
    return {
  File "/Users/lucas/miniforge3/envs/pdod/lib/python3.10/site-packages/mmda/predictors/hf_predictors/vila_predictor.py", line 179, in <dictcomp>
    key: torch.tensor(val, dtype=torch.int64, device=self.device)
TypeError: 'float' object cannot be interpreted as an integer
```

This PR adds an explicit cast operation to get around this issue during pre-processing.

### DIFF
--- a/mmda/predictors/heuristic_predictors/sentence_boundary_predictor.py
+++ b/mmda/predictors/heuristic_predictors/sentence_boundary_predictor.py
@@ -143,6 +143,6 @@ class PysbdSentenceBoundaryPredictor(BaseHeuristicPredictor):
                 itertools.chain.from_iterable([ele.spans for ele in cur_spans])
             )
 
-            sentence_spans.append(SpanGroup(merge_neighbor_spans(all_token_spans)))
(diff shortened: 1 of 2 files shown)


lolipopshock:
Thanks @soldni ! This issue is also fixed in vila https://github.com/allenai/vila/commit/0a772032ef5c31a152d4d849be0c3825acdb1583 and I'll migrate the changes to mmda very soon!

lolipopshock:
fixed in #89 as latest vila lib fixes this issue

PR #80, opened 2022-05-18
reorganize

yoganandc:
i did when i thought i had to work here.


cmwilhelm (changes requested):
This is fine with me, but you'll need to change the pointers in `config.yaml`. They're based on python package paths, i.e. `ai2_internal.timo.vila.interface.Predictor`.

cmwilhelm (approved):
👌

PR #83, opened 2022-06-17
Bib Entry Parser/Predictor

stefanc-ai2:
https://github.com/allenai/scholar/issues/32461

Pretty standard model and interface implementation. You can see the result on http://bibentry-predictor.v0.dev.models.s2.allenai.org/ or 
```
curl -X 'POST' \
  'http://bibentry-predictor.v0.dev.models.s2.allenai.org/invocations' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "instances": [
    {
      "bib_entry": "[4] Wei Zhuo, Qianyi Zhan, Yuan Liu, Zhenping Xie, and Jing Lu. Context attention heterogeneous network embed- ding. Computational Intelligence and Neuroscience , 2019. doi: 10.1155/2019/8106073."
    }
  ]
}'
```

Tests: integration test passed and dev deployment works

TODO: release as version `0.0.10` after merge

### DIFF
--- a/ai2_internal/config.yaml
+++ b/ai2_internal/config.yaml
@@ -64,4 +64,39 @@ model_variants:
     # Leave this unset unless your model has special system requirements beyond
     # those in your setup.py.
     docker_run_commands: ["apt-get update", "apt-get install python3-opencv -y"]
+  bibentry_parser:
+    # Class path to pydantic Instance implementation in <model_package_name>==<model_package_version>
+    instance: ai2_internal.bibentry_predictor.interface.Instance
 
+    # Class path to pydantic Prediction implementation in <model_package_name>==<model_package_version>
+    prediction: ai2_internal.bibentry_predictor.interface.Prediction
+
+    # Class path to Predictor implementation in <model_package_name>==<model_package_version>
+    predictor: ai2_internal.bibentry_predictor.interface.Predictor
+
+    # Class path to pydantic PredictorConfig implementation in <model_package_name>==<model_package_version>
+    predictor_config: ai2_internal.bibentry_predictor.interface.PredictorConfig
+
+    # Any additional sets of dependencies required by the model.
+    # These are the 'extras_require' keys in your setup.py.
+    extras_require: ["dev", "bibentry_predictor"]
+
+    # Full S3 path to tar.gz'ed artifacts archive, nullable
+    artifacts_s3_path: s3://ai2-timo-registry/model-artifacts/bibentryparser/v0.tar.gz
+
+    # Version of python required for model runtime, e.g. 3.7, 3.8, 3.9
+    python_version: 3.7
+
+    # Whether this model supports CUDA GPU acceleration
+    cuda: False
+
+    # Python path to a fn in <model_package_name>==<model_package_version> that
+    # returns a unittest.TestCase. Builder function receives a model container
+    # as its sole argument.
+    # Used by the TIMO toolchain to validate your model implementation and configuration.
+    integration_test: ai2_internal.bibentry_predictor.integration_test.TestInterfaceIntegration
+
+    # One or more bash commands to execute as part of a RUN step in a Dockerfile.
+    # Leave this unset unless your model has special system requirements beyond
+    # those in your setup.py.
+    docker_run_commands: []
--- a/ai2_internal/config.yaml
+++ b/ai2_internal/config.yaml
@@ -64,7 +64,7 @@ model_variants:
     # Leave this unset unless your model has special system requirements beyond
     # those in your setup.py.
     docker_run_commands: ["apt-get update", "apt-get install python3-opencv -y"]
-  bibentry_parser:
+  bibentry_predictor:
     # Class path to pydantic Instance implementation in <model_package_name>==<model_package_version>
     instance: ai2_internal.bibentry_predictor.interface.Instance
 
--- a/mmda/predictors/hf_predictors/bibentry_predictor.py
+++ b/mmda/predictors/hf_predictors/bibentry_predictor.py
@@ -1,0 +1,217 @@
+from enum import Enum
+import re
+from typing import Dict, List, Optional
+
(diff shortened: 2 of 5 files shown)


mmda/predictors/hf_predictors/bibentry_predictor.py
```diff
@@ -0,0 +1,217 @@
...
+    citation_number: Optional[StringWithSpan]
+    authors: Optional[List[StringWithSpan]]
+    title: Optional[StringWithSpan]
+    journal_venue_or_event: Optional[StringWithSpan]
+    year: Optional[StringWithSpan]
+    doi: Optional[StringWithSpan]
+    url: Optional[StringWithSpan]
+
+
+class BibEntryPredictor:
```
stefanc-ai2, line 51: I've been calling it BibEntryParser, but since it's under `hf_predictor` I renamed it to `BibEntryPredictor`. Which name do you prefer `BibEntryParser` or `BibEntryPredictor`?

mmda/predictors/hf_predictors/bibentry_predictor.py
```diff
@@ -0,0 +1,217 @@
...
+        # and this function should have argstring: def predict(self, document: Document) -> List[Annotation]:
+        # and some processing that looks like:
+        # for bib_entry in document.bib_entries:
+        #     bib_entry.text -> treat this as text
+        res = []
+
+        tokenized_inputs = self.tokenizer(bib_entries, padding=True, truncation=True, return_tensors="pt")
+        predictions = self.model(**tokenized_inputs)
+
+        pred_ids = predictions.logits.argmax(2).tolist()  # check GPU vs CPU; might need to .detach()
```
stefanc-ai2, line 68: Kyle mentioned to `check GPU vs CPU, might need .detach()`. I checked that it works fine on both.

mmda/predictors/hf_predictors/bibentry_predictor.py
```diff
@@ -0,0 +1,217 @@
...
+        title = BibEntryPredictor._clean_str(pred.title.content) if pred.title else None
+        doi = BibEntryPredictor._clean_doi(pred.doi.content) if pred.doi else None
+        return dict(
+            citation_number=citation_number,
+            title=title,
+            doi=doi
+        )
+
+    @staticmethod
+    def _get_word_level_prediction(word_ids: List[Optional[int]], predictions: List[int]) -> List[int]:
```
stefanc-ai2, line 95: Token manipulation gymnastics from here down.

ai2_internal/config.yaml
```diff
@@ -64,4 +64,39 @@ model_variants:
...
+    extras_require: ["dev", "bibentry_predictor"]
+
+    # Full S3 path to tar.gz'ed artifacts archive, nullable
+    artifacts_s3_path: s3://ai2-timo-registry/model-artifacts/bibentryparser/v0.tar.gz
+
+    # Version of python required for model runtime, e.g. 3.7, 3.8, 3.9
+    python_version: 3.7
+
+    # Whether this model supports CUDA GPU acceleration
+    cuda: False
```
stefanc-ai2, line 91: Not sure about this one. It currently works on CPU-only instance, but might not be optimal.

cmwilhelm (commented):
The TIMO-ification aspects here look good to me, thanks for working through that!

The part that I'd like to get more clarity on before we merge is what we talked about in sprint planning this meeting: whether we're going to use the mmda data model.

As far as I'm aware, all the existing mmda predictors use a shared data model to represent documents and their components. I don't fully understand what the ramifications are for NOT using them for the bib entry predictor, other than that it complicates integration in SPP.

@kyleclo @lolipopshock what is your guidance on this front? Should all MMDA models be using the MMDA data model? And if so, how should it be applied to bib entry?

kyleclo:
@cmwilhelm Not using the MMDA model likely results in a few things:
1. MMDA is undergoing some refactoring, and it's easier to refactor if all the Predictors adhere to the same input/output than if each Predictor was doing something on its own.
2. MMDA has a standard way of performing serialization/instantiation to/from JSON.  Deviating from the underlying data model likely means SPP (or whatever downstream user of MMDA) will need to implement/maintain their own serialization.
3. MMDA abstracts away some annoying boilerplate stuff one would have to write to obtain bounding boxes of output of Predictors.  For example, if we ever want to highlight BibEntry titles/authors/DOIs for whatever reason & we needed the exact bounding boxes of these predictions, the code you'd have to write to obtain this from the other MMDA objects is probably harder to write than if the predictor had adhered originally to the MMDA conventions.

For Stefan's specific instance, none of these are particularly severe/important; just a bit of hassle later for #1 and #2. 
 #3 seems unlikely (there are no plans for bounding boxes on the fields that Stefan is extracting) and even if it comes to it later, it's not really hard to adapt.  
 
In fact, #3 is an interesting one because, Stefan's model doesn't *have* to be part of MMDA.  The value of MMDA is mostly to handle a lot of annoying code one would write to handle text + visual annotations on a PDF simultaneously.  Citation & BibEntry highlights, for example, really benefit from this library.  But models like TLDRs or SPECTER, while they *could* conform to MMDA conventions, don't have to as they don't benefit from anything MMDA provides.  One can view Stefan's model producing a JSON of metadata fields given a BibEntry string as falling within the same category of model as TLDRs (give me some text, I'll spit out some other text), as opposed as the category of model for VILA (give me a PDF, I'll spit out the locations of the highlighted units).

In all, I think it's not a big deal if Stefan's model doesn't adhere to MMDA conventions.  In fact, it's probably even fine for now if Stefan's model lives outside of MMDA on its own, similar to TLDRs or SPECTER, until there's a real reason that it *has* to be in MMDA.

Thoughts?

stefanc-ai2:
Merging in this PR. Tracking conversion to taking in `Document` as input and outputting `Annotation` in this ticket. https://github.com/allenai/scholar/issues/32722

PR #86, opened 2022-06-30
Pins vila to 0.3.0.

cmwilhelm:
Unbounded upper version was pulling in breaking
changes transitively in builds.


yoganandc (approved):
@lolipopshock @kyleclo

PR #92, opened 2022-07-08
Change vila lower bound to 0.4.2

lolipopshock:


cmwilhelm (changes requested):
We need to bump `mmda`'s version too for this change to be picked up.

cmwilhelm (approved):

PR #95, opened 2022-07-20
use set membership instead of list membership

cmwilhelm:
Deduplication logic was causing big slow-down in
real world deployments of MMDA models.

The current changeset uses object id to do set
membership checks instead of list membership checks.

Using production S2 data, this resulted in an approximately 400-fold
reduction in `indexers.py` latency for VILA, with identical
output to prior implementation.

Kudos to @Whattabatt both for the profiler tool
that identified the offending line and noticing the `not in`
was likely a linear operation.

Screen Shot 2022-07-20 at 5 08 26 PM

Screen Shot 2022-07-20 at 5 08 11 PM

### DIFF
--- a/mmda/types/indexers.py
+++ b/mmda/types/indexers.py
@@ -34,13 +34,14 @@ class SpanGroupIndexer(Indexer):
         if not isinstance(query, SpanGroup):
             raise ValueError(f'SpanGroupIndexer only works with `query` that is SpanGroup type')
 
-        all_matched_span_groups = []
+        all_matched_span_groups = dict()
+
         for span in query.spans:
             for matched_span_group in self._index[span.start : span.end]:
-                if matched_span_group.data not in all_matched_span_groups: # Deduplicate
-                    all_matched_span_groups.append(matched_span_group.data)
-        # retrieval can be out of order, so sort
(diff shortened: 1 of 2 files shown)


rodneykinney (approved):

Whattabatt (approved):
LGTM!

mmda/types/indexers.py
```diff
@@ -34,13 +34,14 @@ def find(self, query: SpanGroup) -> List[SpanGroup]:
...
         for span in query.spans:
             for matched_span_group in self._index[span.start : span.end]:
-                if matched_span_group.data not in all_matched_span_groups: # Deduplicate
-                    all_matched_span_groups.append(matched_span_group.data)
-        # retrieval can be out of order, so sort
-        return sorted(all_matched_span_groups)
+                object_id = id(matched_span_group.data)
+                all_matched_span_groups[object_id] = matched_span_group.data
+
+        return sorted(list(all_matched_span_groups.values()))
```
kyleclo, line 44: can keep the comment `# retrieval can be out of order, so sort`, else lgtm!

kyleclo (approved):
lgtm; 1 comment

PR #111, opened 2022-07-28
run eval on init

yoganandc:


cmwilhelm (approved):
Maybe a comment inline about why? Since we were both confused by this at one point.

PR #93, opened 2022-07-11
mention predictor

yoganandc:
i got bored waiting for data generation scripts to finish

PR #114, opened 2022-07-29
add citation link predictor to mmda

regan-huff:
addresses https://github.com/allenai/scholar/issues/32685
follows the pattern in https://github.com/allenai/mmda/pull/93 for citation mention predictor
and the cues in the data model notes [here](https://github.com/allenai/scholar/issues/32860)

ran tt verify to run integration test and tt serve-local to verify local swagger docs

### DIFF
--- a/ai2_internal/citation_links/interface.py
+++ b/ai2_internal/citation_links/interface.py
@@ -1,0 +1,81 @@
+"""
+This file contains the classes required by Semantic Scholar's
+TIMO tooling.
+
+You must provide a wrapper around your model, as well
+as a definition of the objects it expects, and those it returns.
+"""
+
+from typing import List, Tuple
+
+from pydantic import BaseModel, BaseSettings
+
+from mmda.predictors.xgb_predictors.citation_link_predictor import CitationLinkPredictor
+from mmda.types import api
+from mmda.types.document import Document
+
+# these should represent the extracted citation mentions and bibliography entries for a paper
+class Instance(BaseModel):
+    """
+    Describes one Instance over which the model performs inference.
+    """
+    mentions: List[api.SpanGroup]
+    bibs: List[api.SpanGroup]
+
+
+class Prediction(BaseModel):
+    """
+    Describes the outcome of inference for one Instance
+    """
+    linked_mentions: List[Tuple[api.SpanGroup, str]]
+
+
+class PredictorConfig(BaseSettings):
+    """
+    Configuration required by the model to do its work.
+    Uninitialized fields will be set via Environment variables.
+    """
+    pass
+
+
+class Predictor:
+    """
+    Interface on to your underlying model.
+
+    This class is instantiated at application startup as a singleton.
+    You should initialize your model inside of it, and implement
+    prediction methods.
+
+    If you specified an artifacts.tar.gz for your model, it will
+    have been extracted to `artifacts_dir`, provided as a constructor
+    arg below.
+    """
+    def __init__(self, config: PredictorConfig, artifacts_dir: str):
+        self._predictor = CitationLinkPredictor(artifacts_dir)
+
+    def predict_one(self, inst: Instance) -> Prediction:
+        """
+        Should produce a single Prediction for the provided Instance.
+        Leverage your underlying model to perform this inference.
+        """
+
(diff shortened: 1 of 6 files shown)


yoganandc (commented):
`mmda.types.api` (and pydantic) was a compromise kyle and i arrived at when we were deploying mmda models from the spp repo. 

now that we have this `ai2_internal` package floating in the repo alongside mmda, i think we can move `mmda/types/api.py` to `ai2_internal` and limit its use (and pydantic) to within there.

the code under `mmda*` generally expects to use the regular (more flexible but no pydantic) data models defined in `mmda/types/{annotation,document,span,box}.py`. this is so that a user can annotate a document with the predictions of several models. for example see https://github.com/allenai/mmda/blob/fbb64a160010b21f9b0f0c87ec279b1d0e075b91/examples/title_abstract.py#L19-L33

also i know you probably followed what was on the ticket but I was thinking more of what we have to store in the annotation store. I feel like the predictor here could just return a list of tuples (mention-id, bib-id)

so the predict signature might be more like
```
def predict(self, doc: Document) -> List[(str,str)]
```
where the Document being passed should already contain mentions and bibs.

breaking it down, let's say in ai2_internal you define some pydantic Instance like so ([ref](https://github.com/allenai/mmda/blob/fbb64a160010b21f9b0f0c87ec279b1d0e075b91/ai2_internal/citation_mentions/interface.py#L18-L25))
```
class Instance(BaseModel):
    """
    Describes one Instance over which the model performs inference.
    """
    symbols: str
    mentions: List[api.SpanGroup]
    bibs: List[api.SpanGroup]
```

you would then convert it to the mmda Document like so
([ref](https://github.com/allenai/mmda/blob/fbb64a160010b21f9b0f0c87ec279b1d0e075b91/ai2_internal/citation_mentions/interface.py#L58-L69))
```
doc = Document(symbols=inst.symbols)
doc.annotate(mentions=[sg.to_mmda() for sg in inst.mentions])
doc.annotate(bibs=[sg.to_mmda() for sg in inst.bibs])
```

this way (as a researcher who is not bothered with ai2_ internal) i can do
```
doc = PDFPlumber().parse(doc)  # <- parser gives symbols, tokens, rows, pages
doc.annotatate(mentions=MentionPredictor().predict(doc))
doc.annotate(bibs=BibPredictor().predict(doc))

links = LinksPredictor().predict(doc)
```

## Why do I need `symbols: str`?

the mmda models/parsers dont actually store text in the annotations even though a property called `text` exists 🤷 

```
class CitationLink(BaseModel):
    mention: api.SpanGroup
    bib: api.SpanGroup

    def to_text_dict(self) -> Dict[str, str]:
        return {"source_text": self.mention.text, "target_text": self.bib.text}
```
is going to give you blank strings for values

here's how to get the text
```
class CitationLink:
    mention: SpanGroup
    bib: SpanGroup

    def to_text_dict(self) -> Dict[str, str]:
        return {"source_text": "".join(mention.symbols), "target_text": "".join(bib.text)}
```
these regular SpanGroups (not pydantic) have a property called `symbols` which returns the text by looking up `symbols: str`. yes, spangroups that have been annotated on a document maintain a reference to the document.

https://github.com/allenai/mmda/blob/fbb64a160010b21f9b0f0c87ec279b1d0e075b91/mmda/types/annotation.py#L117-L119

and you would make all the pairs like so:
```
for mention in doc.mentions:
   for bib in doc.bibs:
      ....
```

ai2_internal/citation_links/interface.py
```diff
@@ -0,0 +1,88 @@
...
+    symbols: str
+    mentions: List[api.SpanGroup]
+    bibs: List[api.SpanGroup]
+
+
+class Prediction(BaseModel):
+    """
+    Describes the outcome of inference for one Instance
+    """
+    linked_mentions: List[Tuple[api.SpanGroup, str]]
```
yoganandc, line 31: suggestion: we just return List[(str,str)] here too? we can make that structure back on the client.

yoganandc (approved):
looks amazing! ⭐

ai2_internal/citation_links/interface.py
regan-huff: works for me!

PR #119, opened 2022-08-05
adds `metadata` field to `Annotation`, failover for `text` property of `SpanGroup`

### LINKED ISSUE #116: Refactor: Better handling of SpanGroup.text vs SpanGroup.symbols
See this PR https://github.com/allenai/mmda/pull/115 for example of trying to stitch the two together.  

We need something that's a bit more baked-into SpanGroups. For example, possibly `SpanGroup.text` is a property that wraps `SpanGroup.symbols` and postproceses using a particular `.join()` logic, but it can be overwritten with data, when available?

soldni:
This PR makes two significant changes:

### (1) Introduce `metadata` field to `Annotation` and all its derived classes

`metadata` is a dictionary designed to hold properties such as the `id` of an Annotation, its `type`, and so own. `metadata` should supersede all previous fields that specify metadata-like information, but that raises a question: how to provide backward compatibility for fields like `id`? 

Enter `@store_field_in_metadata`: this decorator stores information previously specified in fields in the `metadata` dictionary, attaching getters and setters to the class to get and set the value of the field. I recommend reading the docstring of `store_field_in_metadata` to learn more about how it functions.

### (2) Change how `text` attribute of `SpanGroup` works. 

This essentially fixes #116 : `.text` now relies on symbols (that is: `" ".join(span_group.symbols)`) when it has not been set. 

To accomplish this, we actually rely on another useful functionality of `store_field_in_metadata`: providing custom getters and setter for a field. The next `text` attribute is actually stored in the `metadata` field, but, rather that a simple lookup, we build it on-the-fly from symbols in case `text` hasn't been set.


soldni:
rebased over `main` branch, added support to access metadata attributes with dot notation, e.g. 

```python
from mmda.types.annotation import Annotation
a = Annotation()
a.metadata.foo = 'bar'
print(a)     # prints  Annotation(uuid=..., doc=None, metadata=Metadata({'foo': 'bar'}))
```

mmda/types/annotation.py
```diff
@@ -97,7 +126,9 @@ def __getitem__(self, key: int):
 
     def __deepcopy__(self, memo):
         box_group = BoxGroup(
-            id=self.id, boxes=deepcopy(self.boxes, memo), type=self.type, uuid=self.uuid
+            boxes=deepcopy(self.boxes, memo),
```
kyleclo, line 129: a lot of performance issues were resolved by removing deepcopy from other places where annotation() is happening; we may not want to introduce back.  is this critical?

mmda/types/annotation.py
```diff
@@ -2,18 +2,23 @@
...
 from dataclasses import dataclass, field
-from typing import Dict, Iterable, List, Optional
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Union
 from uuid import uuid4
 
 from mmda.types.box import Box
+from mmda.types.metadata import Metadata, store_field_in_metadata
 from mmda.types.span import Span
 
+if TYPE_CHECKING:
```
kyleclo, line 19: wats happening here?

soldni: You can't import `Document` from `mmda.types.document` during code execution because it will result into a circular import, but it's useful to have it around for static type checking. The solution is to use [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING), which gets set to True by type checkers like `mypy`, but is false otherwise.

mmda/types/annotation.py
soldni: This just implements support for `deepcopy`; as long as no code calls it over a span group/box group, we should be good.  

We could perhaps raise a warning here that lets user know that there's a perf penalty associated with it?

mmda/types/annotation.py
kyleclo: didnt know that :O cool

mmda/types/annotation.py
kyleclo: ah gotcha;  yea a warning could be good for now

kyleclo (approved):
lgtm; just add warning for deepcopy

mmda/types/annotation.py
soldni: added!

PR #132, opened 2022-09-03
Kylel/2022 08/smashed

kyleclo:
somewhat hefty PR.  adds an implementation of a general-ish purpose wrapper around HF token classification models.  should be one-stop shop for all models like VILA, BibParsing, MentionDetection, etc. as long as those weights are loadable into a AutoTokenClassification model via from_pretrained().

the implementation here makes heavy use of Smashed, which is a new dependency that takes care of a lot of boilerplate around how to slice up Lists of Dictionaries into formats that are required for a HF model input.

the tests show 2 ways this model can be used: Run over entire slices of a Document, or Run over subsets of a Document.  The former is demonstrated via defining `context` to `pages` and classifying every token.  We verify that no token goes un-labeled.   The latter is demonstrated via defining `context` to `bibs` which is a subset of the full Document; we classify all tokens within those `bibs` and verify that they can be annotated properly onto the original Document with minimal bespoke code to keep track of indices.


setup.py
```diff
@@ -20,6 +20,7 @@
         "dev": ["pytest"],
         "spacy_predictors": ["spacy"],
         "lp_predictors": ["layoutparser", "torch", "torchvision", "effdet"],
+        "hf_predictors": ["torch", "transformers", "smashed@git+https://github.com/allenai/smashed@v0.0.1"],
```
soldni, line 23: You can change this to `smashed==0.1.10` to install from PyPI instead.

mmda/predictors/hf_predictors/span_group_classification_predictor.py
```diff
@@ -0,0 +1,358 @@
...
+        tokenizer = transformers.AutoTokenizer.from_pretrained(
+            pretrained_model_name_or_path=model_name_or_path, *args, **kwargs
+        )
+        config = transformers.AutoConfig.from_pretrained(
+            pretrained_model_name_or_path=model_name_or_path, *args, **kwargs
+        )
+        model = transformers.AutoModelForTokenClassification.from_pretrained(
+            pretrained_model_name_or_path=model_name_or_path, *args, **kwargs
+        )
+        predictor = cls(model, config, tokenizer,
```
soldni, line 180: I would use kwargs for everything, i.e., 

```suggestion
        predictor = cls(model=model, config=config, tokenizer=tokenizer,
```

setup.py
soldni, line 23: You can change this to `smashed==0.1.10` to install from PyPI instead.

mmda/predictors/hf_predictors/span_group_classification_predictor.py
```diff
@@ -0,0 +1,358 @@
...
+    _HF_RESERVED_WORD_IDS = 'word_ids'
+    _HF_RESERVED_WORD_PAD_VALUE = -1
+
+    def __init__(
+            self,
+            model: Any,
+            config: Any,
+            tokenizer: Any,
+            span_group_name: str,
+            context_name: str,
```
soldni, line 95: not something to fix, but we really ought to come up with a standard name for contexts/annotations/groups/???  It took me way to long to figure out what this was

mmda/predictors/hf_predictors/span_group_classification_predictor.py
```diff
@@ -0,0 +1,358 @@
...
+            fields_pad_ids={
+                self._HF_RESERVED_WORD_IDS: self._HF_RESERVED_WORD_PAD_VALUE
+            }
+        )
+        # this casts python Dict[List] into tensors.  if using GPU, would do `device='gpu'`
+        self.python_to_torch_mapper = Python2TorchMapper(
+            device=device
+        )
+        # combining everything
+        self.preprocess_mapper = self.tokenizer_mapper >> \
```
soldni, line 150: 😎🆒

setup.py
kyleclo: oku

mmda/predictors/hf_predictors/span_group_classification_predictor.py
kyleclo: done

soldni (approved):
LGTM now!

PR #134, opened 2022-09-07
angelez/more specific on bib detector dependencies

geli-gel:
Confirmed tt verify with -g works with this new setup and config -- going to see if this fixes cuda error on deploy (this is also what i did in extract-stuff-from-pdfs setup.py): https://stackoverflow.com/a/66728637

error:
image


cmwilhelm (approved):
Let's give it a whirl!

geli-gel:
It's working! 🎉

PR #135, opened 2022-09-07
Kylel/2022 09/span group utils

kyleclo:
minor PR: added tests for a pretty important functionality that was being used -- how to combine Spans that are next to each other into a single big Span.  the key thing that was undocumented was that Boxes for the underlying Spans actually disappear after this merging functionality, which the tests now capture.

dont think this is intended behavior we want to support in future, but for now, this is how this utility is being used

### DIFF
--- a/mmda/utils/tools.py
+++ b/mmda/utils/tools.py
@@ -1,10 +1,9 @@
-from typing import List, Union, Dict, Any, Tuple
+from typing import List, Union, Dict, Any, Tuple, Optional
 
 from mmda.types.span import Span
-from mmda.types.box import Box
 
 
-def merge_neighbor_spans(spans: List[Span], distance) -> List[Span]:
+def merge_neighbor_spans(spans: List[Span], distance: Optional[int]) -> List[Span]:
     """Merge neighboring spans in a list of un-overlapped spans:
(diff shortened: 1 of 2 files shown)


cmwilhelm (approved):
This makes sense to me as a way of documenting existing behavior and catching behavioral regressions if/when changes are made to mmda in the future.

Question though -- this code path gets invoked when one does `document.annotate(fieldname=[some, boxgroups])`, right? So you're saying that in the automatic translation to spangroups, we will outright lose bbox information despite that being the form of the information originally?

If so, that strikes me likely as being a bug rather than something we're actively depending on right now. I suppose we can at least recover the bboxes doing something like `document.fieldname.tokens.boxes`?

geli-gel:
@cmwilhelm It looks like we don't lose original bbox information because it's still there in the SpanGroup's BoxGroup

kyleclo:
@geli-gel 's right. the only place this is lost is in the `Span`'s `Box`, not in the `SpanGroup`'s `BoxGroup`.  

i dont think this is a bug; it's just poor/opaque design.  ive created an issue to visit this later: https://github.com/allenai/mmda/issues/137

PR #138, opened 2022-09-08
Kyle/2022 09/pdfplumber symbols

kyleclo:
this PR handles 2 things:

1. make default usage of Pdfplumber predictor to be our current most recommended setting (the setting that is most conducive for citation mention detection). we want everyone to be developing their models off as similar pdf tokens data as possible.

2. slight modification to citation mention detection model to import its split punctuation variable from pdfplumber parser itself; minimize risk that they deviate over time


cmwilhelm (approved):
Make sense to me. So now, when we upgrade the pdf plumber service, instead of instantiating it with:

```
PdfPlumber(split_at_punctuation="magiccharacterset")
```

we just do:

```
PdfPlumber()
```

correct?

cmwilhelm (changes requested):
We should bump the version in `setup.py` so we can publish and use this code.

kyleclo:
thx; made the bump in setup

that's right @cmwilhelm , we'd generally be asking MMDA users to use our default setting & specify their own splitting strategy at their own risk. 

there are downsides to this in that changes to this default are less visible, but that's up to the users of the library to decide if they want to maintain `magicstring` on their end

cmwilhelm (approved):
I think it's probably fine as long as we consider changes to the default string a "breaking" change that induces an appropriate version bump.

PR #141, opened 2022-09-22
quality of life improvements related to importing mmda types, predictors, etc.

kyleclo:
1. make `from mmda.types|predictors|parsers import *` possible
2. migrate api to ai2 internal


mmda/predictors/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+__all__ = ['DictionaryWordPredictor']
+
+with necessary('pysbd', soft=True) as PYSBD_AVAILABLE:
+    if PYSBD_AVAILABLE:
+        from mmda.predictors.heuristic_predictors.sentence_boundary_predictor \
+            import PysbdSentenceBoundaryPredictor
+        __all__.append('PysbdSentenceBoundaryPredictor')
+
+with necessary(["layoutparser", "torch", "torchvision", "effdet"], soft=True) as PYTORCH_AVAILABLE:
```
cmwilhelm, line 14: It would be nice to not have to define this in three places:

1. setup extras_require
2. required_backends on the predictor itself
3. here

But I don't have a good suggestion.

cmwilhelm (approved):
I think this all makes sense. You'll want to bump setup.py version +0.0.1 again because I think @geli-gel beat you to the punch with v0.0.36.

Putting the api types into ai2_internal also seems like a good move but note it will be breaking for the parsers/models still operationalized in spp instead of TIMO. Should be relatively easy to fix if/when we upgrade them though, I think.

mmda/predictors/__init__.py
kyleclo: yeah :/ ive created an issue here: https://github.com/allenai/mmda/issues/144

i need to think a bit more how this would be resolved. a reasonable stop-gap could be defining a top-level config that all 3 locations imports from; reserve for next PR

PR #142, opened 2022-09-22
Bib and mention ids

geli-gel:
https://github.com/allenai/scholar/issues/33626
fixes bib and mention models to have unique ids per pdf rather than per page

todo:
- [ ] tt publish both and update versions in timo-services


cmwilhelm (approved):
LGTM Angele, thanks for hunting this one down

PR #143, opened 2022-09-23
Bib detect span groups

geli-gel:
Part of integrating the linker model: https://github.com/allenai/scholar/issues/33591
The linker model accepts SpanGroups
to avoid complicating the linker model and requiring parser parts to generate span groups (and avoid drift between spangroups created between models that might have different versions of mmda) this PR updates the bib detection model to generate and output SpanGroups instead of BoxGroups (and BoxGroups are conserved within those SpanGroups).

TODO: 
- [x] version up in setup.py before merging this
- [ ] update timo-services

Goes hand in hand with this PR updating SPP: https://github.com/allenai/spp/pull/127


ai2_internal/bibentry_detection_predictor/interface.py
```diff
@@ -99,8 +99,11 @@ def predict_one(self, inst: Instance) -> Prediction:
 
         processed_bib_entry_box_groups, original_box_groups = self._predictor.predict(doc, self._config.BIB_ENTRY_DETECTION_MIN_VILA_BIB_ROWS)
 
+        # generate SpanGroups
+        doc.annotate(bib_entries=processed_bib_entry_box_groups)
```
cmwilhelm, line 103: sneaky sneaky

cmwilhelm (approved):
LGTM. However, note that if we update the timo service in place with this, it will break SPP in dev/local, since those envs depend on the bib_entry_boxes existing as fields on the response object.

PR #146, opened 2022-09-28
Egork/vila 0.5

comorado:


setup.py
```diff
@@ -23,7 +23,7 @@
         "pysbd_predictors": ["pysbd"],
         "lp_predictors": ["layoutparser", "torch", "torchvision", "effdet"],
         "hf_predictors": ["torch", "transformers", "smashed==0.1.10"],
-        "vila_predictors": ["vila>=0.4.2,<0.5", "transformers"],
+        "vila_predictors": ["vila>=0.5", "transformers"],
```
cmwilhelm, line 26: Let's maintain an upper bound on the version. Previously we've had changes in behavior because only the floor was pinned and vila.git introduced breaking behavior.

```["vila>0.5,<0.6",```

cmwilhelm (changes requested):
Let's set an upper bound on the vila dependency per inline comment.

Let's also bump mmda's version in `setup.py` so you can publish it as an update via `tt publish`.

cmwilhelm (approved):

PR #147, opened 2022-09-28
Allows empty annotations.

cmwilhelm:
Previously we would skip making attrs
on the document to represent these empty
annotation groups. This resulted in attr
not found errors in contexts where an expected
field was absent and legitimately empty.


kyleclo (approved):
lgtm

PR #148, opened 2022-09-29
set python version for bib predictor to 3.8

cmwilhelm:


cmwilhelm:
ci is failing on this because we're over our limit on retrieving artifacts from HF i think

geli-gel (approved):
TY

PR #150, opened 2022-09-30
Add metadata to API data classes

soldni:
This PR adds metadata to API data classes. 

API data classes and mmda types differ in a few significant aspects:

- `id` and `type` (and `text` for `SpanGroup`)  are stored in `metadata` for mmda types; in the APIs, they are part of the top-level attributes. 
- `metadata` can store arbitrary content in the mmda types; in the data API, all attributes that are not explicitly declared are dropped. 
    - we expect applications that require specific fields to declare custom Metadata and SpanGroup/BoxGroup classes that inherit from their parent class. For an example, see tests/test_internal_ai2/test_api.py

### DIFF
--- a/ai2_internal/api.py
+++ b/ai2_internal/api.py
@@ -1,10 +1,8 @@
-from __future__ import annotations
+from typing import List, Optional, Type
 
-from typing import List, Optional
 from pydantic import BaseModel
 
 import mmda.types.annotation as mmda_ann
-
 
 __all__ = ["BoxGroup", "SpanGroup"]
 
@@ -17,14 +15,18 @@ class Box(BaseModel):
     page: int
 
     @classmethod
-    def from_mmda(cls, box: mmda_ann.Box) -> Box:
+    def from_mmda(cls, box: mmda_ann.Box) -> "Box":
         return cls(
             left=box.l, top=box.t, width=box.w, height=box.h, page=box.page
         )
 
     def to_mmda(self) -> mmda_ann.Box:
         return mmda_ann.Box(
-            l=self.left, t=self.top, w=self.width, h=self.height, page=self.page
+            l=self.left,
+            t=self.top,
+            w=self.width,
+            h=self.height,
+            page=self.page
         )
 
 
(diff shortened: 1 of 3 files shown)


soldni:
@cmwilhelm per @yoganandc suggestion, I'm now mapping what is the `metadata` field in MMDA to the `attributes` field in the Annotation Store. 

This should better fit the design of the annotation store. All previous considerations regarding attributes being typed are still valid. 

Let me know what you think!

yoganandc (commented):
i was hoping for something much simpler...

can we just have a
```
class Annotation(BaseModel):
  id: Optional[str]
  type: Optional[str]
  attributes: Dict[str, Any] = {}

class BoxGroup(Annotation):
class SpanGroup(Annotation):
```

and then in `to_mmda()`
```
pydantic_annotation.attributes = dict(mmda_annotation.metadata)
```

soldni:
> i was hoping for something much simpler...
> 
> can we just have a
> 
> ```
> class Annotation(BaseModel):
>   id: Optional[str]
>   type: Optional[str]
>   attributes: Dict[str, Any] = {}
> 
> class BoxGroup(Annotation):
> class SpanGroup(Annotation):
> ```
> 
> and then in `to_mmda()`
> 
> ```
> pydantic_annotation.attributes = dict(mmda_annotation.metadata)
> ```

The tricky part in this is that, in `mmda`, we made the decision to store `id` and `type` in the metadata. in fact, `mmda_annotation.id`, `mmda_annotation.type`, and `mmda_annotation.text` are just aliases to `mmda_annotation.metadata.id`,  `mmda_annotation.metadata.type`, `mmda_annotation.metadata.text` (where aliases are actually implemented through descriptors). 

The reason for such decision was that nor `id` nor `type` are actually required in `mmda`, so we moved off the top level but maintained aliases for backward compatibility. So, when converting to a pydantic object that conforms to SPP APIs, we have to remove them from metadata.

yoganandc:
> So, when converting to a pydantic object that conforms to SPP APIs, we have to remove them from metadata.

👍 a little more logic but still fits the overall structure im proposing right?

cmwilhelm:
@yoganandc the complexity here I think is @Lucaweihs responding to my direct request to have some sort of document schema for however a given model is using metadata. If model authors will be dumping principal inference data into them, we need a way to specify what the shape of that data is. Otherwise no one is going to be able to make sense of it later.

yoganandc (approved):
works for me then. @cmwilhelm maybe you have more thoughts since you're working on the client PR?

ai2_internal/api.py
```diff
@@ -34,60 +36,106 @@ class Span(BaseModel):
...
+class Annotation(BaseModel, extra=Extra.ignore):
+    attributes: Attributes
+
+    @classmethod
+    def get_metadata_cls(cls) -> Type[Attributes]:
+        for inherit_cls in cls.mro():
+            if not hasattr(inherit_cls, "__annotations__"):
+                continue
+            if "attributes" in inherit_cls.__annotations__:
+                return inherit_cls.__annotations__["attributes"]
```
cmwilhelm, line 75: I think we could do this via:

```python
@classmethod
def get_metadata_cls(cls) -> Type[Attributes]:
    return cls.__fields__["attributes"]["type"]
```

rather than using the low level python api and traversing the hierarchy.

cmwilhelm (approved):
This LGTM but I had one small suggestion.

We'll also need to bump version in setup.py again because we've claimed 0.0.43 :)

ai2_internal/api.py
soldni: done!

ai2_internal/api.py
```diff
@@ -17,14 +16,18 @@ class Box(BaseModel):
...
     @classmethod
-    def from_mmda(cls, box: mmda_ann.Box) -> Box:
+    def from_mmda(cls, box: mmda_ann.Box) -> "Box":
         return cls(
             left=box.l, top=box.t, width=box.w, height=box.h, page=box.page
         )
 
     def to_mmda(self) -> mmda_ann.Box:
         return mmda_ann.Box(
-            l=self.left, t=self.top, w=self.width, h=self.height, page=self.page
```
comorado, line 27: @soldni it seems that this change is breaking the behavior of this function https://github.com/allenai/mmda/blob/main/mmda/types/box.py#L78

Do we need to update the Box class then?

PR #158, opened 2022-10-11
[wip] adding relations

kyleclo:


mmda/types/annotation.py
```diff
@@ -42,54 +59,72 @@ def __init__(
...
+        objects of another field. For example `page.tokens` jumps from a particular page
+        to all its intersecting tokens."""
+        if not self.doc:
+            raise AttributeError("This annotation does not have an attached document")
 
         if field in self.doc.fields:
             return self.doc.find_overlapping(self, field)
 
+        # TODO[kylel] - when does this ever get called? infinite loop?
         return self.__getattribute__(field)
```
soldni, line 111: You probably can remove it; [this stack overflow post][1] has a good explanation, but tl;dr is that `__getattribute__` supersedes `__getattr__` if defined, so you will never be in a situation where `__getattribute__`  should be called after `__getattr__`. 

I suspect that the function of this call here is to raise an error if `field` is not defined; if that is the case, I would explicitly raise an `AttributeError` instead with a more descriptive error message. 

[1]: https://stackoverflow.com/a/3278104/938048

mmda/types/annotation.py
```diff
@@ -42,54 +59,72 @@ def __init__(
...
 
-
 class BoxGroup(Annotation):
     def __init__(
             self,
             boxes: List[Box],
             id: Optional[int] = None,
             doc: Optional['Document'] = None,
+            field: Optional[str] = None,
             metadata: Optional[Metadata] = None,
```
soldni, line 121: Not related to this PR specifically, but I think we should document these arguments in a docstring.

mmda/types/document.py
```diff
@@ -245,16 +244,23 @@ def from_json(cls, doc_dict: Dict) -> "Document":
...
-                field_name_to_span_groups[field_name] = span_groups
+                field_to_span_groups[field] = span_groups
 
         # 3) load annotations for each field
-        doc.annotate(**field_name_to_span_groups)
+        doc.annotate(**field_to_span_groups)
 
         return doc
+
+    def locate_annotation(self, name: AnnotationName) -> Annotation:
```
soldni, line 261: I am not quite sure on why we need `AnnotationName`. If a name is just a combo of relation name + id, can we just expect a tuple of the two here? or require two args? Overall, having a class to just hold annotation names would just result in computational overhead.

soldni (commented):
The overall design seems good to me! I don't quite understand why we need `AnnotationName` classes though. What does the extra overhead of this class get us?

mmda/types/annotation.py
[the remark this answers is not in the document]
kyleclo: fixed

kyleclo:
@soldni 
> The overall design seems good to me! I don't quite understand why we need AnnotationName classes though. What does the extra overhead of this class get us?

Without the class, we would need to code somewhere how IDs are constructed in the library. For now, it's `field_name - integer_id`, but it's possible in the future this will need to be extended. 

As well, we need some way to parse this ID for use in lookup of that specific element within a Document. I don't want `field, id = obj.split('-')` everywhere throughout the code as it gets hard to maintain in case we ever change something. The class allows us to have methods `.field` and `.id` for use here.

mmda/types/annotation.py
kyleclo: removed

mmda/types/annotation.py
kyleclo: yea we'll need that

mmda/types/document.py
kyleclo: see comment below

mmda/types/annotation.py
```diff
@@ -34,6 +35,22 @@ def warn_deepcopy_of_annotation(obj: "Annotation") -> None:
...
+        self.id = id
+
+    def __str__(self) -> str:
+        return f"{self.field}-{self.id}"
+
+    @classmethod
+    def from_str(cls, s: str) -> 'AnnotationName':
+        field, id = s.split('-')
+        id = int(id)
+        return AnnotationName(field=field, id=id)
```
soldni, line 52:
```suggestion
        return cls(field=field, id=id)
```

Prevents issues when inheriting (if we ever decide to).

mmda/types/annotation.py
```diff
@@ -34,6 +35,22 @@ def warn_deepcopy_of_annotation(obj: "Annotation") -> None:
     warnings.warn(msg, UserWarning, stacklevel=2)
 
 
+class AnnotationName:
+    """Stores a name that uniquely identifies this Annotation within a Document"""
+
```
soldni, line 40:
```suggestion
__slots__ = ("field", "id")

```

Speeds up class creation by roughly 20%:

```python
>>> %timeit AnnotationName('relation', 1)
140 ns ± 0.391 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

# after adding slots

>>> %timeit AnnotationName('relation', 1)
115 ns ± 0.163 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
````

soldni (approved):

soldni:
> @soldni
> 
> > The overall design seems good to me! I don't quite understand why we need AnnotationName classes though. What does the extra overhead of this class get us?
> 
> Without the class, we would need to code somewhere how IDs are constructed in the library. For now, it's `field_name - integer_id`, but it's possible in the future this will need to be extended.
> 
> As well, we need some way to parse this ID for use in lookup of that specific element within a Document. I don't want `field, id = obj.split('-')` everywhere throughout the code as it gets hard to maintain in case we ever change something. The class allows us to have methods `.field` and `.id` for use here.

@kyleclo Sounds good! added two small suggestions to improve it, but otherwise ok to merge!

PR #166, opened 2022-10-19
Kylel/2022 10/hotfix spacing in pdfplumber symbols

kyleclo:
Before, due to tokenization from PDF Plumber that would split on every punctuation, we were observing Documents contained symbols that looked like:
```
>> doc.symbols:
References Mitchell P . Marcus , Beatrice Santorini , and Mary Ann Marcinkiewicz . 1993 .

>> doc.tokens:
References
Mitchell
P
.
Marcus
,
Beatrice
Santorini
,
and
Mary
Ann
Marcinkiewicz
.
1993
.
```

This is because the way PDFPlumberParser constructs `.symbols` to include in Document was previously using either whitespace `" "` or newline `"\n"` concatenation of the strings in each detected token.

What we actually want is to keep this fine-grained partitioning of tokens, but to still have `.symbols` better preserve the  (lack of) whitespacing that exists in the original document, when appropriate. Ideally:

```
>> doc.symbols:
References Mitchell P. Marcus, Beatrice Santorini, and Mary Ann Marcinkiewicz. 1993.

>> doc.tokens:
References
Mitchell
P
.
Marcus
,
Beatrice
Santorini
,
and
Mary
Ann
Marcinkiewicz
.
1993
.
```

This fix is a reimplementation of the logic of how `PDFPlumberParser` pulls information out of `PDFPlumber` objects for stitching together.  The implementation is a bit tricky to explain, but see the new tests to get a sense of functionality.

### DIFF
--- a/mmda/parsers/pdfplumber_parser.py
+++ b/mmda/parsers/pdfplumber_parser.py
@@ -99,60 +99,13 @@ class PDFPlumberParser(Parser):
         self.split_at_punctuation = split_at_punctuation
 
     def parse(self, input_pdf_path: str) -> Document:
-        doc = self._load_pdf_as_doc(input_pdf_path)
-        return doc
-
-    def _load_page_tokens(
-        self,
-        page: pdfplumber.page.Page,
-        page_index: int,
-        x_tolerance: int = 1.5,
-        y_tolerance: int = 2,
-        keep_blank_chars: bool = False,
-        use_text_flow: bool = True,
-        horizontal_ltr: bool = True,
-        vertical_ttb: bool = True,
-        extra_attrs: Optional[List[str]] = None,
-        split_at_punctuation: Optional[Union[str, bool]] = None,
-    ):
-        token_data = page.extract_words(
-            x_tolerance=x_tolerance,
-            y_tolerance=y_tolerance,
-            keep_blank_chars=keep_blank_chars,
-            use_text_flow=use_text_flow,
-            horizontal_ltr=horizontal_ltr,
-            vertical_ttb=vertical_ttb,
-            extra_attrs=extra_attrs,
-            split_at_punctuation=split_at_punctuation
-        )
-        page_tokens = [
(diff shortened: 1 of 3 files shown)


mmda/parsers/pdfplumber_parser.py
```diff
@@ -99,171 +100,200 @@ def __init__(
...
+            coarse_tokens = page.extract_words(
+                x_tolerance=self.token_x_tolerance,
+                y_tolerance=self.token_y_tolerance,
+                keep_blank_chars=self.keep_blank_chars,
+                use_text_flow=self.use_text_flow,
+                horizontal_ltr=self.horizontal_ltr,
+                vertical_ttb=self.vertical_ttb,
+                extra_attrs=self.extra_attrs,
+                split_at_punctuation=None
+            )
```
geli-gel, line 121: is it "ok" resource-wise that we're gonna do `.extract_words` twice? Is this the part of PDFPlumber that takes a long time?

geli-gel (commented):
Thanks for doing this! Ok - my understanding is, this change in how doc.symbols is stored is what will enable DWP to go on working as before because it does this: 
https://github.com/allenai/mmda/blob/bd3a0a9466699122ba4abee58649bea1bebb72dc/mmda/predictors/heuristic_predictors/dictionary_word_predictor.py#L186-L197
correct?

mmda/parsers/pdfplumber_parser.py
kyleclo: i'm not sure if it's OK or not. but this is an acceptable fix for now. we can always profile & try a separate implementation in future if it seems like this is too inefficient, but my impression now is that the biggest gains in efficiency arent here but in the predictors

kyleclo:
@geli-gel good call on DictionaryWordPredictor. ill take a look at that PR, but i dont want this one to have dependency on that predictor, as this is more core-functionality

PR #169, opened 2022-10-19
Egork/vila unitet test to class

comorado:
Moving unite tests to the class. Locally pytest test would fail on this file

### DIFF
--- a/.github/workflows/mmda-ci.yml
+++ b/.github/workflows/mmda-ci.yml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Test with Python ${{ matrix.python-version }}
         run: |
-          pip install -e .[dev,pysbd_predictors,hf_predictors]
+          pip install -e .[dev,test,pysbd_predictors,hf_predictors]
           pytest tests --ignore=tests/test_predictors/test_vila_predictors.py --ignore=tests/test_predictors/test_figure_table_predictors.py
 
   test_vila_predictors:
@@ -39,7 +39,7 @@ jobs:
           sudo apt-get update
           sudo apt-get -y install poppler-utils
       - run: |
-          pip install -e .[dev]
+          pip install -e .[dev,test]
           pip install -e .[vila_predictors]
           pytest tests/test_predictors/test_vila_predictors.py
           pytest tests/test_predictors/test_figure_table_predictors.py
--- a/.github/workflows/mmda-ci.yml
+++ b/.github/workflows/mmda-ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Test with Python ${{ matrix.python-version }}
         run: |
           pip install -e .[dev,pysbd_predictors,hf_predictors]
-          pytest tests --ignore=tests/test_predictors/test_vila_predictors.py --ignore=tests/test_predictors/test_figure_table_predictors.py
+          pytest --cov-fail-under=42 --ignore=tests/test_predictors/test_vila_predictors.py --ignore=tests/test_predictors/test_figure_table_predictors.py
 
   test_vila_predictors:
     runs-on: ubuntu-latest
@@ -41,5 +41,5 @@ jobs:
       - run: |
           pip install -e .[dev]
           pip install -e .[vila_predictors]
-          pytest tests/test_predictors/test_vila_predictors.py
-          pytest tests/test_predictors/test_figure_table_predictors.py
+          pytest tests/test_predictors/test_vila_predictors.py --no-cov
+          pytest tests/test_predictors/test_figure_table_predictors.py --no-cov
--- a/.github/workflows/mmda-ci.yml
+++ b/.github/workflows/mmda-ci.yml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Test with Python ${{ matrix.python-version }}
         run: |
-          pip install -e .[dev,test,pysbd_predictors,hf_predictors]
+          pip install -e .[dev,pysbd_predictors,hf_predictors]
           pytest tests --ignore=tests/test_predictors/test_vila_predictors.py --ignore=tests/test_predictors/test_figure_table_predictors.py
 
   test_vila_predictors:
@@ -39,7 +39,7 @@ jobs:
           sudo apt-get update
           sudo apt-get -y install poppler-utils
       - run: |
-          pip install -e .[dev,test]
+          pip install -e .[dev]
           pip install -e .[vila_predictors]
           pytest tests/test_predictors/test_vila_predictors.py
           pytest tests/test_predictors/test_figure_table_predictors.py
--- a/README.md
+++ b/README.md
@@ -9,6 +9,19 @@ conda create -n mmda python=3.8
 pip install -e '.[dev,<extras_require section from setup.py>]'
 ```
 
+## Unit testing
+```bash
+pytest ./tests
+```
+for latest failed test
+```bash
+pytest ./tests --lf
+```
+for specific test name of class name
+```bash
+pytest ./tests -k <'name_of_class_or_unit_test'>
+```
+
 ## Quickstart guide
 
 #### 1. Create a Document for the first time from a PDF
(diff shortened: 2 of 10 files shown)


comorado:
> Hey @comorado, I am not sure if it recommended to move directories during testing: I can imagine introducing a bug during tests that is caused by a previous test moving to a different directory.
> 
Good point @soldni, that's the reason why I had it in setUp method. setUp is run prior to running each individual test https://stackoverflow.com/questions/6854658/explain-the-setup-and-teardown-python-methods-used-in-test-cases


> In tests I've authored, I use `__file__` to get an absolute path to fixtures, e.g. see [`tests/test_parsers/test_override.py`](https://github.com/allenai/mmda/blob/c7cb515333b877d8c4a5d4e88aec81960a250874/tests/test_parsers/test_override.py#L13-L16).
I like your suggestion and instead of changing the directories I will create a class variable which points to the fixtures path.
> 
> Some [StackOverflow answers](https://stackoverflow.com/a/71617139/938048) recommend creating functions pointing to paths and decorating them with `@pytest.fixture`

A bit too fancy, thank you for the pointer :)

comorado:
> > Hey @comorado, I am not sure if it recommended to move directories during testing: I can imagine introducing a bug during tests that is caused by a previous test moving to a different directory.
> 
> Good point @soldni, that's the reason why I had it in setUp method. setUp is run prior to running each individual test https://stackoverflow.com/questions/6854658/explain-the-setup-and-teardown-python-methods-used-in-test-cases
> 
> > In tests I've authored, I use `__file__` to get an absolute path to fixtures, e.g. see [`tests/test_parsers/test_override.py`](https://github.com/allenai/mmda/blob/c7cb515333b877d8c4a5d4e88aec81960a250874/tests/test_parsers/test_override.py#L13-L16).
> > I like your suggestion and instead of changing the directories I will create a class variable which points to the fixtures path.
> > Some [StackOverflow answers](https://stackoverflow.com/a/71617139/938048) recommend creating functions pointing to paths and decorating them with `@pytest.fixture`
> 
> A bit too fancy, thank you for the pointer :)

@soldni I've pushed a commit where I am creating a path variable. Still doing it though in the setUp method. On the pytest run time each individual test receives it's version of the setUp variables. So for example if self.doc was modified in one test, another test is not going to see the modifications because it receives its version of the self.doc created by the setUp class.

README.md
```diff
@@ -9,6 +9,19 @@ conda create -n mmda python=3.8
...
+```bash
+pytest ./tests
+```
+for latest failed test
+```bash
+pytest ./tests --lf
+```
+for specific test name of class name
+```bash
+pytest ./tests -k <'name_of_class_or_unit_test'>
```
geli-gel, line 22: Maybe an example of how a working command would look exactly?

geli-gel (approved):

README.md
comorado: Sure thing, I've also added pytest.ini and packages for runing tests on multiple cpus.

.github/workflows/mmda-ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Test with Python ${{ matrix.python-version }}
         run: |
-          pip install -e .[dev,pysbd_predictors,hf_predictors]
+          pip install -e .[dev,test,pysbd_predictors,hf_predictors]
```
soldni, line 26: Could you roll `test` dependencies into `dev` dependencies? is not like someone is going to do dev work on MMDA without running tests 😜

tests/test_parsers/test_pdf_plumber_parser.py
```diff
@@ -7,14 +7,14 @@
...
 
 class TestPDFPlumberParser(unittest.TestCase):
-    
+    def setUp(cls) -> None:
+        cls.fixture_path = os.path.join(pathlib.Path(__file__).parent.parent, 'fixtures')
+
     def test_parse(self):
         parser = PDFPlumberParser()
-        doc = parser.parse(input_pdf_path="../fixtures/1903.10676.pdf")
+        doc = parser.parse(input_pdf_path=os.path.join(self.fixture_path, '1903.10676.pdf'))
```
soldni, line 17:
```suggestion
        doc = parser.parse(input_pdf_path=self.fixture_path / '1903.10676.pdf')
```

tests/test_predictors/test_figure_table_predictors.py
```diff
@@ -14,15 +14,17 @@
...
     @classmethod
     def setUp(cls):
+        cls.fixture_path = pathlib.Path(__file__).parent.parent
         cls.doc = (PDFPlumberParser()
-                   .parse(input_pdf_path='fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf'))
-        cls.images = PDF2ImageRasterizer().rasterize(input_pdf_path='fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf',
+                   .parse(input_pdf_path=os.path.join(cls.fixture_path,
+                                                      'fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf')))
+        cls.images = PDF2ImageRasterizer().rasterize(input_pdf_path=os.path.join(
+            cls.fixture_path, 'fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf'),
```
soldni, line 27:
```suggestion
                   .parse(input_pdf_path=cls.fixture_path, / 'fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf'))
        cls.images = PDF2ImageRasterizer().rasterize(input_pdf_path=cls.fixture_path, / 'fixtures/0c027af0ee9c1901c57f6579d903aedee7f4.pdf',
```

tests/test_predictors/test_vila_predictors.py
```diff
@@ -15,130 +16,130 @@
...
+
+    def test_vila_predictors(self):
+        layout_predictor = LayoutParserPredictor.from_pretrained(
+            "lp://efficientdet/PubLayNet"
+        )
+
+        pdfplumber_parser = PDFPlumberParser()
+        rasterizer = PDF2ImageRasterizer()
+
+        doc = pdfplumber_parser.parse(input_pdf_path=os.path.join(self.fixture_path, "1903.10676.pdf"))
```
soldni, line 70:
```suggestion
        doc = pdfplumber_parser.parse(input_pdf_path=self.fixture_path / "1903.10676.pdf")
```

tests/test_predictors/test_vila_predictors.py
```diff
@@ -15,130 +16,130 @@
...
+        images = rasterizer.rasterize(input_pdf_path=os.path.join(self.fixture_path, "1903.10676.pdf"), dpi=72)
```
soldni, line 71:
```suggestion
        images = rasterizer.rasterize(input_pdf_path=self.fixture_path / "1903.10676.pdf", dpi=72)
```

tests/test_predictors/test_vila_predictors.py
```diff
@@ -15,130 +16,130 @@
...
+
+        resB = ivilaB.predict(doc)
+        del ivilaB
+
+        assert [ele.spans for ele in resA] == [ele.spans for ele in resB]
+        assert [ele.type for ele in resA] == [self.S2VL_LABEL_MAP[ele.type] for ele in resB]
+
+    def test_vila_predictors_with_special_unicode_inputs(self):
+
+        test_doc_path = os.path.join(self.fixture_path, "unicode-test.json")
```
soldni, line 133:
```suggestion
        test_doc_path = self.fixture_path / "unicode-test.json"
```

soldni (commented):
I don't know if I found them all, but suggested a few places where you can use `pathlib`'s `/` syntax to join paths. From 

```python
import os
from pathlib import Path

path = Path('path/to/dir')
os.path.join(path, 'subpath/to/file.txt')
```

to
```python
from pathlib import Path

path = Path('path/to/dir')
path / 'subpath/to/file.txt'
```

saves you an import and it's easier to read!

--------

I've also left a comment about unifying `dev` and `test` dependency groups.

soldni (approved):
lgtm!

PR #172, opened 2022-10-25
Egork/project.toml

comorado:


comorado:
@soldni Brought to my attention new suggest way of configuring [setuptools using pyproject.toml file](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#configuring-setuptools-using-pyproject-toml-files)

In this PR I am proposing following changes:
- pyproject.toml instead of setup.py, we still keeping minimal setup.py for backward compatibility (timo).
- move mmda source code and ai2_internals down into src folder for less of the confusion. 

I

soldni (approved):

PR #176, opened 2022-11-14
Fixes citation linker and errata

cmwilhelm:
sklearn is no longer a real package on pypi
and instead was a deprecation timebomb waiting to blow up.
See:
https://pypi.org/project/sklearn/

Must now install as "scikit-learn" for reasons.

* separately provides a default to some new parameter that was added on
  `api.SapnGroup`, as we're not explicitly passing this

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.2.1'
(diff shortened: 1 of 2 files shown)


regan-huff (approved):
ooh fun
sorry about that

PR #177, opened 2022-11-15
[bug] Fix dangling file reference leading to ResourceWarning

rauthur:
In exceptional cases this pdfplumber reference can be left open. Addresses ResourceWarning observed in tests.


kyleclo (approved):
lgtm

PR #182, opened 2022-11-29
Add fontinfo to tokens without requiring word split

rauthur:
This PR appends font information (font name and size) as metadata to 'tokens' on a Document without requiring tokens to be split on that information (i.e., "best" effort if a token contains many font names or sizes). The code subclasses the WordExtractor provided by PDFPlumber.

Currently, in a default configuration of `PDFPlumberParser`, tokens are already extracted with font name and size information (although it is discarded and only used for splitting). This could be added to the metadata as-is, however, the method used is the `extra_attrs` argument of PDFPlumber's `extract_words` which forces token splitting if font name and size does not match. I believe this is a bad default and further is not required (users can override this argument). The approach here guarantees this metadata will be captured.

Re: "best effort" in case of many name/size options for a token: I have maintained the logic of just taking the font name and size from the first character in the token. Another approach provides the min, max and average font sizes (or others) as well as a set of font names. This could be provided in the future or by adapting the `append_attrs` argument. For current use cases (section nesting prediction) this approach is sufficient.


src/mmda/parsers/pdfplumber_parser.py
```diff
@@ -1,15 +1,56 @@
...
+from mmda.types.span import Span
+
+
+class WordExtractorWithFontInfo(ppu.WordExtractor):
+    """Override WordExtractor methods to append additional char-level info."""
+
+    def __init__(
+        self,
+        x_tolerance: T_num = ppu.DEFAULT_X_TOLERANCE,
+        y_tolerance: T_num = ppu.DEFAULT_Y_TOLERANCE,
```
soldni, line 24: T_num is `int`, right? we could annotate with that and avoid the risky import from the private class.

src/mmda/parsers/pdfplumber_parser.py
```diff
@@ -1,15 +1,56 @@
...
+        super().__init__(
+            x_tolerance,
+            y_tolerance,
+            keep_blank_chars,
+            use_text_flow,
+            horizontal_ltr,
+            vertical_ttb,
+            extra_attrs,
+            split_at_punctuation,
+        )
```
soldni, line 42:
```suggestion
        super().__init__(
            x_tolerance=x_tolerance,
            y_tolerance=y_tolerance,
            keep_blank_chars=keep_blank_chars,
            use_text_flow=use_text_flow,
            horizontal_ltr=horizontal_ltr,
            vertical_ttb=vertical_ttb,
            extra_attrs=extra_attrs,
            split_at_punctuation=split_at_punctuation,
        )
```

I would switch to kwargs in case the order changes in pdfplumber.

src/mmda/parsers/pdfplumber_parser.py
```diff
@@ -1,15 +1,56 @@
...
+        self.append_attrs = append_attrs
+
+    def merge_chars(self, ordered_chars: ppu.T_obj_list) -> ppu.T_obj:
+        """Modify returned word to capture additional char-level info."""
+        word = super().merge_chars(ordered_chars=ordered_chars)
+
+        for key in self.append_attrs:
+            if key not in word:
+                word[key] = ordered_chars[0][key]
```
soldni, line 51: maybe add a comment about what is the shape of `ordered_chars`?

src/mmda/parsers/pdfplumber_parser.py
```diff
@@ -179,17 +223,17 @@ def parse(self, input_pdf_path: str) -> Document:
...
             return doc
 
     def _convert_nested_text_to_doc_json(
-            self,
-            token_dicts: List[Dict],
-            word_ids: List[int],
-            row_ids: List[int],
-            page_ids: List[int]
+        self,
+        token_dicts: List[Dict],
```
soldni, line 233:
```suggestion
        token_dicts: List[dict],
```

soldni (approved):
Minor comments, but otherwise looks good!

src/mmda/parsers/pdfplumber_parser.py
rauthur: 👍

PR #184, opened 2022-12-08
Kylel/2022 12/bugfix overlapping spans pdfplumber

kyleclo:
Addressing problems in https://github.com/allenai/scholar/issues/34648. Both PDFs run fine now after this fix + previous version upgrade.

Also upgrade `necessary` which something changed in upstream dependency.


stefanc-ai2 (approved):
Thanks, Kyle!

PR #194, opened 2023-01-23
Return only spanful groups from BibDetector model

geli-gel:


geli-gel:
next steps:
- [ ] tt push
- [ ] update timo-services config for bib-predictor to use new code which will trigger new deployment

src/mmda/predictors/d2_predictors/bibentry_detection_predictor.py
```diff
@@ -120,13 +119,13 @@ def postprocess(self,
...
-                BoxGroup(
-                    boxes=[model_output_box],
-                    id=current_id,
-                    type="raw_model_prediction"
-                )
+            box_group = BoxGroup(
+                boxes=[model_output_box],
+                id=current_id
             )
+            box_group.metadata.set('type', 'raw_model_prediction')
```
regan-huff, line 126: was this change related to a breaking change in how metadata and type are defined for BoxGroup? or do both ways of setting "type" work and this one is more forward-looking for how we want to define metadata from now on?

src/mmda/predictors/__init__.py
```diff
@@ -1,10 +1,12 @@
 # flake8: noqa
 from necessary import necessary
 
-from mmda.predictors.heuristic_predictors.whitespace_predictor import WhitespacePredictor
-from mmda.predictors.heuristic_predictors.dictionary_word_predictor import DictionaryWordPredictor
+with necessary(["tokenizers"], soft=True) as TOKENIZERS_AVAILABLE:
```
regan-huff, line 4: in what case would we run this without the tokenizers?

src/ai2_internal/bibentry_detection_predictor/interface.py
```diff
@@ -98,13 +98,18 @@ def predict_one(self, inst: Instance) -> Prediction:
...
+            self._config.BIB_ENTRY_DETECTION_MIN_VILA_BIB_ROWS
+        )
 
         # generate SpanGroups
         if len(processed_bib_entry_box_groups) > 0:
             doc.annotate(bib_entries=processed_bib_entry_box_groups)
             prediction = Prediction(
-                bib_entries=[api.SpanGroup.from_mmda(sg) for sg in doc.bib_entries],
+                # filter out span-less SpanGroups
+                bib_entries=[api.SpanGroup.from_mmda(sg) for sg in doc.bib_entries if len(sg.spans) != 0],
```
regan-huff, line 111: can you remind me why we end up producing spanless SpanGroups that we then have to filter out?

ok, from your comments above, I remember that we end up with overlapping bib boxes, and then the spans can only be assigned to one of them. But do you have a sense of what causes us to draw the bib boxes badly?

src/ai2_internal/bibentry_detection_predictor/integration_test.py
```diff
@@ -125,3 +126,37 @@ def test__no_resulting_bibs(self, container):
 
         self.assertEqual(len(predictions[0].bib_entries), 0)
         self.assertEqual(len(predictions[0].raw_bib_entry_boxes), 0)
+
+    def test__spanless_bibs(self, container):
```
regan-huff, line 130: nice job testing the behavior!

regan-huff (approved):
Looks good! Just a few questions for my own understanding...

src/mmda/predictors/d2_predictors/bibentry_detection_predictor.py
geli-gel: The first - this is related to a breaking change in MMDA, BoxGroup no longer has "type" directly on it, it has to be part of metadata

src/mmda/predictors/__init__.py
geli-gel: The Bib Detector model doesn't need the 'tokenizers' library, the reason I added this in is because when I ran tt verify at first, it errored because DictionaryWordPredictor required 'tokenizers' - but the Bib Detector model doesn't need DictionaryWordPredictor in the first place!

src/ai2_internal/bibentry_detection_predictor/interface.py
geli-gel: My sense of what causes us to draw the bib boxes badly is not a good understanding but - I know it has to do with the original model output boxes simply being "bad" - it's possible that increasing the threshold of model score for boxes returned by the model might reduce the amount that this happens but I haven't looked closely at enough examples to know for sure if that would help or make things worse by giving us fewer boxes overall - I'm sure there's a sweet spot but I'm not sure how to find it! Simply filtering these ones out seems like the best thing to do for now.

regan-huff: 👍

src/mmda/predictors/d2_predictors/bibentry_detection_predictor.py
regan-huff: Gotcha, I take it this is one of the grab bag of changes you get to pick up with a new version whether you want them all or not!

PR #197, opened 2023-02-23
Ensures compatibility with pdfplumber 0.8.0 and refreshed example

soldni:
Summary of changes:

- PDFPlumber version `0.8.0` deprecated `pdfplumber.utils` in favor of `pdfplumbers.utils.text`. This PR makes import compatible with both previous version and current version of PDF plumber.
- Refreshed example of scientific doc parser to use newer VILA models.


rauthur (approved):
LGTM!

PR #199, opened 2023-03-06
Cap PDFPlumber version

kyleclo:
cap pdfplumber version since they made 0.8.0 release which moves Word…Extractor to a submodule.

Can see here it's now under a submodule `text`: https://github.com/jsvine/pdfplumber/blob/b6847ad4cd4e54e201c0301f66b2c1f3e914cdc0/pdfplumber/utils/text.py

which didn't exist before in 0.7.6 or earlier versions. We were relying on the WordExtractor from: https://github.com/jsvine/pdfplumber/blob/v0.7.6/pdfplumber/utils.py

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ requires-python = '>=3.7'
 dependencies = [
         'tqdm',
         'pdf2image',
-        'pdfplumber>0.7.1',
+        'pdfplumber>0.7.1,0.8',
         'requests',
         'pandas',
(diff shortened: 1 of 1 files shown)


kyleclo:
nvm, cancelling PR, @soldni already fixed it https://github.com/allenai/mmda/blob/065e5681f2617a0e0ca6d19e4ea1694a79b42d9a/src/mmda/parsers/pdfplumber_parser.py#L8

PR #201, opened 2023-03-07
TIMO Citation linker returns Relations objects.

cmwilhelm:
Purely additive change that has the citation linker
timo predictor return links in the future-standard
generalized `Relation` format.

This is the format expected by clients in new SPP
code.

cc also @geli-gel as this may be relevant for grobid
work

### DIFF
--- a/src/ai2_internal/api.py
+++ b/src/ai2_internal/api.py
@@ -158,3 +158,9 @@ class SpanGroup(Annotation):
             box_group=self.box_group.to_mmda()if self.box_group else None,
             id=self.id
         )
+
+
+class Relation(BaseModel):
+    from_id: int
+    to_id: int
+    attributes: Attributes = Attributes()
--- a/src/ai2_internal/citation_links/interface.py
+++ b/src/ai2_internal/citation_links/interface.py
@@ -31,6 +31,7 @@ class Prediction(BaseModel):
     """
     # tuple represents mention.id and bib.id for the linked pair
     linked_mentions: List[Tuple[str, str]]
+    linked_mention_relations: List[api.Relation]
 
 
 class PredictorConfig(BaseSettings):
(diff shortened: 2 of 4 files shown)


src/ai2_internal/api.py
```diff
@@ -158,3 +158,9 @@ def to_mmda(self) -> mmda_ann.SpanGroup:
             box_group=self.box_group.to_mmda()if self.box_group else None,
             id=self.id
         )
+
+
+class Relation(BaseModel):
+    from_id: int
+    to_id: int
+    attributes: Attributes = Attributes()
```
regan-huff, line 166: The thought is that one day the kind of relation will need to be specified here, as an attribute? But for now we leave it unspecified?

regan-huff (approved):
lgtm

src/ai2_internal/citation_links/interface.py
```diff
@@ -65,9 +66,15 @@ def predict_one(self, inst: Instance) -> Prediction:
...
 
-        prediction = self._predictor.predict(doc) # returns (mention.id, bib.id)
-        
-        return Prediction(linked_mentions = prediction)
+        prediction = self._predictor.predict(doc) # returns [(mention.id, bib.id)]
+
+        return Prediction(
+            linked_mentions = prediction,
+            linked_mention_relations = [
+                api.Relation(from_id=mention_id, to_id=bib_id)
```
geli-gel, line 74: I thought we'd have to include the from and to annotation names at the timo interface ("mentions", "bib_entries")

cmwilhelm: I discussed this at length with Tyler and Chloe a couple days ago because they had comments on my SPP PR. I think we decided the TIMO endpoint would just have a named field with no special designation beyond that.

src/ai2_internal/api.py
cmwilhelm: More general than that -- any metadata associated with the relation. Scores, or whatever. Arbitrary bag of data.

Kyle has it in mind that when one annotates an MMDA doc with relations you would specify their identifier, which is a 3-tuple of `(type, sourceAnnoName, targetAnnoName)` -- but that identifier is not in the `Relation` object itself.

PR #204, opened 2023-03-08
Grobid Full Text Parser

geli-gel:
full text grobid parser, works like this (examples/grobid_full_text_parser/grobid_bibs.ipynb):

parser.parse takes in an optional output directory to save the raw grobid xml.
the parser takes in a config set with the basic grobid_client setup + an additionally requested piece "head" which contains the authors section. (had to look in chrome dev tools to find that out because I could only get the authors in the gui but not from the client, and couldn't find the info anywhere, also could not read the grobid js code!)

this is the first part of https://github.com/allenai/scholar/issues/35751 -- second part is going to be taking these raw grobid bibs and updating the spangroups so that the SpanGroup spans include tokens that indicate a bib's number, as well as use the SpanGroup span boxes which essentially consolidate the grobid boxes which are drawn on a per-line basis (whereas the spangroup span boxes are drawn on a per text block basis, more similar to the bib detector we are using grobid in place of)

Next steps for this grobid parser are to make it output more pieces of the grobid output.

image


src/mmda/parsers/grobid_full_text_parser.py
```diff
@@ -0,0 +1,112 @@
...
+        
+        # if/when adding in relations between mention sources and bib targets, we'll have to 
+        #  re-map the source and target IDs onto their new MMDA SpanGroup IDs
+        doc.annotate(
+            bib_entries=bib_entries
+            )
+
+        return doc
+
+    def _xml_coords_to_boxes(self, coords_attribute: str, page_sizes: dict):
```
regan-huff, line 71: is there a way to test that this works as expected? Like, that this conversion from Grobid's output gives us something close to what we get via the mmda bib detector, which (I believe) follows a different path to boxes?

you mention this as work still to come:
>use the SpanGroup span boxes which essentially consolidate the grobid boxes which are drawn on a per-line basis (whereas the spangroup span boxes are drawn on a per text block basis, more similar to the bib detector we are using grobid in place of

I'm not understanding yet how this is something that still needs to be done...can you explain, or annotate the code somehow so it is clearer what step might still be missing? or is there a way a test could clarify what this means?

src/mmda/parsers/grobid_full_text_parser.py
```diff
@@ -0,0 +1,112 @@
+"""
+
+@geli-gel
+
+"""
+from grobid_client.grobid_client import GrobidClient
```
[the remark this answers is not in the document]
geli-gel: oh that's right! let me fix that

src/mmda/parsers/grobid_full_text_parser.py
geli-gel: I meant that as a note in regard to the overarching ticket, that this PR represents basically the first half of the work -- getting the raw grobid bibs into MMDA format - so our work in this repo is essentially finished for the bib detection replacement work. In the spp-grobid-api is where I was imagining the SpanGroup span altering code would live.

geli-gel: > is there a way to test that this works as expected? Like, that this conversion from Grobid's output gives us something close to what we get via the mmda bib detector, which (I believe) follows a different path to boxes?

Once we have this code plus the other stuff in spp-grobid would be the better time to test, but as a super simple test I included a small sample in the python notebook showing the text extracted from the grobid boxes:
image

regan-huff (approved):

tests/test_parsers/test_grobid_full_text_parser.py
```diff
@@ -0,0 +1,51 @@
...
+        def status_code(self):
+            return self._status_code
+
+    return MockResponse(XML_OK, 200)
+
+
+class TestGrobidFullTextParser(unittest.TestCase):
+    @um.patch("requests.request", side_effect=mock_request)
+    def test_processes_full_text(self, mock_request):
+        logging.getLogger("pdfminer").setLevel(logging.WARNING)
```
kyleclo, line 37: wats goin on here?

.github/workflows/mmda-ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Test with Python ${{ matrix.python-version }}
         run: |
-          pip install -e .[dev,pysbd_predictors,hf_predictors,lp_predictors]
+          pip install -e .[dev,pysbd_predictors,hf_predictors,lp_predictors,grobid_full_text_parser]
```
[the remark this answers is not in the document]
kyleclo: ah nvm i see, we dont actually test running grobid itself here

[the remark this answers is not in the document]
geli-gel: the test mocks hitting grobid by just returning the xml that grobid would return so that it doesn't need grobid at all (I copied the method used for the test for the pre-existing grobid parser)

tests/test_parsers/test_grobid_full_text_parser.py
geli-gel: I'm not sure why, but when I ran pytest on this file, a bunch of pdfminer DEBUG logs were popping up, so google told me to do this to turn it down a notch.

src/mmda/parsers/grobid_full_text_parser.py
```diff
@@ -0,0 +1,112 @@
...
+from mmda.parsers.parser import Parser
+from mmda.types.annotation import BoxGroup, Box
+from mmda.types.document import Document
+from mmda.types.names import PagesField, RowsField, TokensField
+
+REQUIRED_DOCUMENT_FIELDS = [PagesField, RowsField, TokensField]
+NS = {"tei": "http://www.tei-c.org/ns/1.0"}
+
+
+class GrobidFullTextParser(Parser):
```
kyleclo, line 22: can we rename this to make it clear in the name that this is an unusual Parser that takes as input also a separate `Document`? like `GrobidAugmentExistingDocumentParser`?  then it wouldnt clash if we add a `GrobidFullTextParser` later which strictly creates new documents from the Xml itself without any mapping

tests/test_parsers/test_grobid_full_text_parser.py
kyleclo: ah gotcha! i think i kno wats happening here, thx for the solution, might actually add this closer to where pdfminer appears

kyleclo (approved):
lgtm; minor note about naming

src/mmda/parsers/grobid_full_text_parser.py
geli-gel: would it make sense to make a new directory under mmda called `augmenters`? and name it `GrobidAugmenter`? just cause i like shorter names

PR #210, opened 2023-03-15
Merge Spans per page when merging spans by symbol distance

geli-gel:
group Spans by pages when merging SpanGroup Spans by symbol distance to avoid "boxes are on different pages" error

This error is coming up when annotating Grobid BoxGroups onto a Document because some BoxGroups contain boxes across pages with spans that might be close together in symbol distance.

related to allenai/scholar/issues/35751

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.3.0'
+version = '0.3.1'
 description = 'MMDA - multimodal document analysis'
 authors = [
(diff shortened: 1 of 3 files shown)


geli-gel:
@kyleclo also passed `tt verify` for `bibentry_detection_predictor`, `ivila-row-layoutlm-finetuned-s2vl-v2`, and `figure_table_predictors`.

kyleclo (approved):
seems fine for now; as we chatted, ill try a diff method that circumvents this altogether

kyleclo:
@geli-gel can u also increment the version in pyproject

PR #211, opened 2023-03-16
Move grobid_id to metadata in grobid augmenter

geli-gel:
was foolish of me to store the grobid id in the id field -- it's not allowed by ai2_internal api


kyleclo (approved):
not sure what u need the grobid ID for, but seems fine to me!

PR #214, opened 2023-03-21
Bib entry predictor handles empty inputs

cmwilhelm:
RE: https://github.com/allenai/scholar/issues/35525

An empty list of bib entries is valid input for
the bib entry predictor. Previously we were blowing up.
Now we return an empty prediction set.


geli-gel (approved):
TY

PR #217, opened 2023-04-05
get spans with _annotate_box_group even if token boxes are on box_group instead of span

geli-gel:
part of https://github.com/allenai/scholar/issues/35862 -- ran into issue where we're unable to determine underlying text from box_groups because SPP annotation store keeps PdfPlumber output in "SpanGroups with BoxGroups" format, so the token box that's expected to be on the token span's `.box` is no longer found there.
This PR updates `_annotate_box_group`  so that we check a page of tokens to determine where the token's box is stored, and pass that information on to `allocate_overlapping_tokens_for_box` so that tool also knows where to check for the boxes, as well as adds a method to `MergeSpans` so that a default `MergeSpans` class with spans can be instantiated with SpanGroups instead (specifically for this use case).

todo:
- [x] add a test (in progress)


src/mmda/utils/tools.py
```diff
@@ -132,12 +169,21 @@ def build_merged_spans_from_connected_components(self, index):
...
+                    merged_box = Box.small_boxes_to_big_box(
+                        [span.box for span in page_spans]
+                    )
+                    merged_spans.append(
+                        Span(
+                            start=min([span.start for span in page_spans]),
+                            end=max([span.end for span in page_spans]),
+                            box=merged_box,
+                        )
+                    )
```
geli-gel, line 188: this is just formatting

src/mmda/types/document.py
```diff
@@ -160,37 +159,61 @@ def _annotate_box_group(
...
-                            for span_group in self.pages[box.page].tokens
+                    cur_page_tokens = all_page_tokens[box.page] = self.pages[
+                        box.page
+                    ].tokens
+                    if token_box_in_box_group is None:
+                        # Determine whether box is stored on token SpanGroup span.box or in the box_group
+                        token_box_in_box_group = all(
+                            [
+                                (
+                                    (hasattr(token.box_group, "boxes") and len(token.box_group.boxes) == 1)
```
cmwilhelm, line 180: I think this might be a little off. `SpanGroup.box_group` is `Optional[BoxGroup]`, and `BoxGroup.boxes` is `List[Box]`.

So this should probably be:

```python
(hasattr(token, "box_group") and len(token.box_group.boxes) == 1)
```

cmwilhelm: If we don't already have it, we could use a test case where we don't have box groups on the tokens, which would throw an exception as written right now.

cmwilhelm (changes requested):
I *think* this makes sense to me, although I'd want @kyleclo 's okay on it first. I did not realize the code was quite so sprawling when I suggested this approach, so thanks for getting into it.

I think in the future we may not preserve the boxes on the merged spans (since `box` will be going away on `Span`), but that's a problem for another day.

I did have one concrete comment about the `token_box_in_box_group` check.

src/mmda/types/document.py
```diff
@@ -160,37 +159,61 @@ def _annotate_box_group(
...
                 if box.page not in all_page_tokens:
-                    cur_page_tokens = all_page_tokens[box.page] = list(
-                        itertools.chain.from_iterable(
-                            span_group.spans
-                            for span_group in self.pages[box.page].tokens
+                    cur_page_tokens = all_page_tokens[box.page] = self.pages[
+                        box.page
+                    ].tokens
+                    if token_box_in_box_group is None:
+                        # Determine whether box is stored on token SpanGroup span.box or in the box_group
```
kyleclo, line 176: can we add a `logging.warning` if they go down the route of `token SpanGroup span.box` that we will deprecate support for this and that they should make sure `Document` is annotated with `tokens` with `Box Groups`

kyleclo (approved):
lgtm - chatted in person; 1 log statement & agree w chrisw's proposed test

cmwilhelm (approved):

src/mmda/types/document.py
cmwilhelm: I think we should do this but wait till we get closer to the mmda refactor work being taken on. This will produce A LOT of logs and might blow up datadog if we let it sit for months on end

PR #219, opened 2023-04-07
New nb old grobid removed

geli-gel:
Fix for https://github.com/allenai/scholar/issues/35862
Old Grobid pieces removed and s2 ID matching working the same way as s2matchmaker (since s2matchmaker doesn't yet add matches back to Annotation Store)
image

This is the notebook we'll want to use for evaluation


baileykuehl (approved):
Lgtm, thanks Angele!

PR #224, opened 2023-04-14
Egork/heuristic approach figure table pred v2

comorado:
Hello @cmwilhelm, it is ready. I will query the service locally to have last checks and quality of predictions. Pleas let me know of your comments meanwhile.

Thank you!


src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
```diff
@@ -110,58 +351,58 @@ def _predict(doc: Document, caption_type: str = 'Figure') -> List[SpanGroup]:
...
                         box=merged_boxes_caption_dict[page][col].box)],
-                        box_group=BoxGroup(boxes=[merged_boxes_fig_dict[page][row].box]),
+                        box_group=BoxGroup(boxes=[merged_boxes_fig_tab_dict[page][row].box]),
                         metadata=Metadata(type=caption_type)
-                    ))
-        return predictions
+                    )
+                    span_group.text = self.doc.symbols[span_group.start: span_group.end]
+                    if span_group.text.lower().startswith(caption_type.lower()[:3]):
+                        span_group.id = hash(json.dumps(span_group.to_json()))
```
cmwilhelm, line 387: I think a simpler approach for generating the ids here is just to make them indices at time of creation. That's what we've done for other models that populate this field, and I think that's the mid-/long-term plan for MMDA, right @kyleclo ?

src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
```diff
@@ -12,15 +13,169 @@
...
 from mmda.types.span import Span
 from mmda.utils.tools import MergeSpans
+from ai2_internal.api import Relation
 
 
 class FigureTablePredictions(BaseHeuristicPredictor):
     """Class for creating a map of figure boxes to figure captions
     """
-    REQUIRED_DOCUMENT_FIELDS = ['pages', 'tokens', 'rows', 'layoutparser_span_groups', 'vila_span_groups']
+    REQUIRED_DOCUMENT_FIELDS = ['pages', 'tokens', 'rows', 'vila_span_groups']
```
cmwilhelm, line 22: You still requires `"block"` don't you?

src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
cmwilhelm: For instance:

```python
box_group.id = len(predictions)
span_group.id = len(predictions_captions)
```

would probably be much cleaner and simpler.

src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
```diff
@@ -110,58 +351,58 @@ def _predict(doc: Document, caption_type: str = 'Figure') -> List[SpanGroup]:
...
+                            fig_box.box, caption_box)
 
                         cost_matrix[j][i] = distance
                         if caption_type == 'Figure':
                             cost_matrix[j][i] = distance if distance > 0 else 900
 
                 row_ind, col_ind = linear_sum_assignment(cost_matrix)
                 for row, col in zip(row_ind, col_ind):
-                    predictions.append(SpanGroup(spans=[Span(
+                    span_group = SpanGroup(spans=[Span(
```
cmwilhelm, line 378: This construction is slightly awkward because you are subsequently pulling the box group out and then adding it to a separate list. At the end of this process, we don't want the SG to have a BG anymore, so ideally we don't put it there in the first place. What about:

```python
for row, col in zip(row_ind, col_ind):
    span_group = SpanGroup(
        spans=[
            Span(
                start=merged_boxes_caption_dict[page][col].start,
                end=merged_boxes_caption_dict[page][col].box
            )
        ]
    )

    box_group = BoxGroup(boxes=[merged_boxes_fig_tab_dict[page][row][box])
```

Note I have done three things:

1. never put the box group on the SG
2. don't put boxes on the individual spans (we don't want/need them there, will break spp persistence and violates how Entities will be constructed in mmda refactor)
3. dropped Metadata from your SG because this was only being used to differentiate between table and figure captions. And you are not saving those as separate lists of spangroups, so they are disambiguated there instead.

src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
```diff
@@ -110,58 +351,58 @@ def _predict(doc: Document, caption_type: str = 'Figure') -> List[SpanGroup]:
...
-        return predictions
+                    )
+                    span_group.text = self.doc.symbols[span_group.start: span_group.end]
+                    if span_group.text.lower().startswith(caption_type.lower()[:3]):
+                        span_group.id = hash(json.dumps(span_group.to_json()))
+                        box_group = span_group.box_group
+                        box_group.id = hash(json.dumps(box_group.to_json()))
+                        predictions.append(box_group)
+                        predictions_captions.append(span_group)
+                        predictions_relations.append(Relation(from_id=span_group.id, to_id=box_group.id))
```
cmwilhelm, line 392: I think encoding the relationship in either direction is fine, since it's 1-1. However, we should be consistent.

The names for your fields are:
"figure_to_figure_captions" and "table_to_table_captions"

but in this code you are making the caption the "from" entity, and the figure/table the "to" entity. Please either reverse your `from_id` and `to_id` assignment here, or rename your fields to:
"figure_caption_to_figures" and "table_captions_to_tables"

cmwilhelm (changes requested):
I had a number of small comments, Egor. Mostly related to construction of the SG and BGs and their value assignments. I think it can be simplified and tidied up a bit.

Also noted a discrepancy in the way the Relations are directed between how fields are named and values are assigned.

src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
comorado: I like a lot more thank you, very neat, hopefully will not have collisions from different predictors span_groups.

comorado: > For instance:
> 
> ```python
> box_group.id = len(predictions)
> span_group.id = len(predictions_captions)
> ```
> 
> would probably be much cleaner and simpler.

I agree as long as we do not need to have those id's unique. I have another use of the hash in the code, but in that case I do need to have ids unique. I am merging spang_groups which overlap and it is a graph based algorithm where I do use this ids and need to have them unique.

cmwilhelm: The future for MMDA is that users will not be able to generate these ids themselves, they will automatically be assigned by MMDA, and they are only guaranteed to be unique within a set of spangroups (i.e. "vila_span_groups" or similar), not globally.

Since manual ID assignment -- especially assignment with the characteristics you are using here -- will eventually go away, I suggest you find another way to uniquely identify them. For instance, serial ID assignment as I suggested, prefixed with the name of the spangroup set would be sufficient, i.e. `f"figure_captions_${caption.id}"`

cmwilhelm (approved):
LGTM. Let's just bump the version of mmda in pyproject.toml as part of this PR so you can `tt publish` this revision after it merges to main.

PR #227, opened 2023-04-21
Egork/quick fix/figure table caption prediction updating blocks type

comorado:
Hello @cmwilhelm here is the PR:

- I've changed interface like we discussed for the blocks input
- I've updated integration_test code and fixtures
- I've added notebook for creating fixture for the integration test
- I've removed images folder which is been unused in the tests
- I've cleaned up the notebook I've previously committed

Thank you!


comorado:
I did run `tt verify` and `tt serve-local` and used that endpoint in producing results in the notebook I've comitted previously

cmwilhelm (approved):
LGTM Egor. Feel free to merge and `tt publish`. Then off to `timo.git` to update the service config.

PR #229, opened 2023-04-22
add fixtures; add test of fixtures

kyleclo:


kyleclo:
@khyathiraghavi for documentation purposes, could you add a `README.md` file with:
- high level description of the final recommended approach, 
- what other ideas you tried & why are they worse than the recommended approach  
- what is the dataset you use to develop the model & why you used this dataset? 
- what's the origin of the blocklist & why you used this blocklist? 

also I'm having a bit of trouble knowing which model to use? it looks like there is:
- `pretrain_data/hate_speech/LRClassifier.py`
- `pretrain_data/hate_speech/hate_speech_detector.py`
- `pretrain_data/hate_speech/hate_speech_svm_model.py`

PR #231, opened 2023-04-26
Egork/quick fix/replacing get

comorado:
Hello @cmwilhelm 

- I've fixed the issue with using .get, replaced it with [] on defaultdict.
- I've removed fields from the requirement list (rows, images), I've also removed them from the fixture.
- I've added a new fixture which corresponds to the document failed in SPP
- I've added some logger timing statements for debugging purposes, can remove them


src/mmda/predictors/heuristic_predictors/figure_table_predictors.py
```diff
@@ -106,7 +104,7 @@ def generate_map_of_layout_to_tokens(
         """
         for page in vila_dict.keys():
             for span in vila_dict[page]:
-                for layout_span in dict_of_pages_layoutparser.get(page):
+                for layout_span in dict_of_pages_layoutparser[page]:
```
cmwilhelm, line 107: This works right now but depends on the method being called with a defaultdict(list), which seems like a regression waiting to happen.

Can we do

```
for layout_span in dict_of_pages_layoutparser.get(page, []):
```

instead, for clarity and future proofing?

cmwilhelm (approved):
Made a small comment inline.

Let's also remove the debug statements as they are for a separate issue.

PR #232, opened 2023-04-27
Bib detector entity like output

geli-gel:
for allenai/scholar#36565
Removes span boxes from bib_entry SpanGroups, and turns raw_bib_entry_boxes from BoxGroups into SpanGroups (with no spans because we just care about saving the boxes here, and calling .annotate with the original box_groups is not necessary).

`tt verify -c src/ai2_internal/config.yaml -v bibentry_detection_predictor` passes

todo:
- [ ] publish new mmda version
- [ ] add it to timo-services config
- [ ] see if SPP is happy with it

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.4.0'
+version = '0.4.1'
 description = 'MMDA - multimodal document analysis'
 authors = [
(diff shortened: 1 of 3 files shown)


cmwilhelm (approved):
LGTM. note: spp will indeed still be pissed about this because of format=BoxGroup here:
https://github.com/allenai/spp/blob/main/src/main/resources/spp/base.conf#L186

PR #236, opened 2023-04-28
Fix vila it and drop `blocks`

cmwilhelm:
I would like to redeploy vila with some new performance
modifications. This PR does two things:

1. Removes the requirements for `blocks` from the
  request payload. We don't actually use them,
  since they we are using ivila. This will save us unnecessary,
  expensive boxgroup annotation time.
2. Updates the test fixtures to use the newer `box` json
  serialization, since this was out of date. (values otherwise unchanged)


geli-gel:
this whole time i thought vila annotated the blocks

geli-gel (approved):
what were the blocks for?! 😵

cmwilhelm:
@geli-gel the blocks are used by the hvila variant, which we decided not to use lololol.

PR #237, opened 2023-04-28
bibdetector - give both outputs ids at SpanGroup level

geli-gel:
follow up to https://github.com/allenai/mmda/pull/232 
after plugging in to SPP discovered citation-linker 500s caused by SpanGroups missing IDs


cmwilhelm (approved):
it's good to remember where you've come from.

PR #240, opened 2023-05-01
E2E eval nb to ai2_internal/evaluation_notebooks

geli-gel:
Eval NB for bib entries pipeline that incorporates changes listed https://github.com/allenai/scholar/issues/36474

Replaces `bailey_annot_branch` so that we can put the notebooks into MMDA main branch

`src/ai2_internal/evaluation_notebooks/end_to_end_eval_easier.ipynb` is the updated notebook for evaluation

`src/ai2_internal/evaluation_notebooks/end_to_end_eval_grobid_comparison.ipynb` is there to preserve our grobid comparison code

to see what the nb looks like: https://github.com/allenai/mmda/blob/new_annot_branch/src/ai2_internal/evaluation_notebooks/end_to_end_eval_easier.ipynb


baileykuehl (approved):

geli-gel:
Last update before merge will be:
- [ ] new versions for new datasets after new mmda `box_groups_to_span_groups` tools is used in each timo model and samples run through

PR #246, opened 2023-05-12
Authors in grobid augmenter

geli-gel:
For https://github.com/allenai/scholar/issues/36779


README.md
```diff
@@ -6,12 +6,13 @@ This is work in progress...
...
 ```bash
 conda create -n mmda python=3.8
+conda activate mmda
 pip install -e '.[dev,<extras_require section from setup.py>]'
 ```
 
 For most users, we recommend using recipes:
 ```python
-pip install -e '.[dev,recipes]'
+pip install -e '.[dev,grobid_augment_existing_document_parser]'
```
geli-gel, line 15: Oops, I’ll revert this

src/mmda/parsers/grobid_augment_existing_document_parser.py
```diff
@@ -61,61 +61,87 @@ def parse(self, input_pdf_path: str, doc: Document, xml_out_dir: Optional[str] =
...
+            boxes = self._xml_coords_to_boxes(coords_string)
             grobid_id = bib.attrib["{http://www.w3.org/XML/1998/namespace}id"]
 
-            grobid_bibs.append(
+            bib_entries.append(
                 BoxGroup(
-                boxes=boxes, 
-                metadata=Metadata(grobid_id=grobid_id)
+                    boxes=boxes,
+                    metadata=Metadata(grobid_id=grobid_id)
```
regan-huff, line 143: It feels like this is the only reason we can't write one `_get_grobid_box_groups` method which just gets passed the needed tei keys for a given element. Do we definitely need this?

src/mmda/parsers/grobid_augment_existing_document_parser.py
```diff
@@ -61,61 +61,87 @@ def parse(self, input_pdf_path: str, doc: Document, xml_out_dir: Optional[str] =
...
+        box_group_parses["authors"] = self._get_author_box_groups(xml_root)
         # note for if/when adding in relations between mention sources and bib targets:
         # big_entries metadata contains original grobid id attached to the BoxGroup.
-        doc.annotate(
-            bib_entries=bib_entry_span_groups
-            )
+        box_group_parses["bib_entries"] = self._get_bib_entry_box_groups(xml_root)
+
+        # get nice SpanGroups from tokens whose centers overlap with the Grobid boxes
+        annotatable_span_groups_by_field_name = self._grobid_box_groups_to_span_groups(doc, **box_group_parses)
```
regan-huff, line 84: this might just be my brain, but I think I would rather go field by field than make this nifty dict that kind of hides what you're actually adding to the doc. Is that possible? 

Something like:
```
author_span_groups = box_groups_to_span_groups(get_author_box_groups(xml_root))
bib_entry_span_groups = box_groups_to_span_groups(get_bib_entry_box_groups(xml_root))
doc.annotate(author_span_groups)
doc.annotate(bib_entry_span_groups)
```

regan-huff (approved):
Looks groovy!

src/mmda/parsers/grobid_augment_existing_document_parser.py
geli-gel: That's a good question ... we don't definitely need this, unless we certainly want to make Grobid mentions and links to bibs available from SPP. If it turns out we do need this, i think there can be ways to add in optional extra metadata from a single `_get_grobid_box_groups` method -- and for now since we aren't even using that id I'm happy to take it out and try refactoring a little

regan-huff (approved):
Thank you for the refactor!

PR #247, opened 2023-05-17
SVM for predicting hyphenated words

kyleclo:


src/mmda/predictors/heuristic_predictors/svm_word_predictor.py
```diff
@@ -0,0 +1,390 @@
+"""
+
+
+"""
```
cmwilhelm, line 4: nit: suggest dropping the empty docstring or fleshing it out

cmwilhelm (approved):
This looks good to my eyes aside from some light comments, and has the hooks I need to wrap with a TIMO interface. I can't do a deep dive on the code, but I'm satisfied if the test set you're using looks good, and we do a round of bug squashing from the SPP pipeline.

src/mmda/types/span.py
```diff
@@ -24,32 +24,32 @@ def to_json(self) -> Dict:
...
     @classmethod
-    def small_spans_to_big_span(cls, spans: List['Span']) -> 'Span':
+    def small_spans_to_big_span(cls, spans: List["Span"]) -> "Span":
         # TODO: add warning for unsorted spans or not-contiguous spans
         # TODO: what happens when Boxes cant be merged?
-        start = None
-        end = None
-        for span in spans:
+        start = spans[0].start
+        end = spans[0].end
```
geli-gel, line 45: was this broken?

tests/test_predictors/test_svm_word_predictor.py
```diff
@@ -0,0 +1,286 @@
...
+                    {"id": 4, "spans": [{"start": 15, "end": 16}]},
+                    {"id": 5, "spans": [{"start": 16, "end": 18}]},
+                    {"id": 6, "spans": [{"start": 18, "end": 19}]},
+                    {"id": 7, "spans": [{"start": 19, "end": 21}]},
+                    {"id": 8, "spans": [{"start": 21, "end": 22}]},
+                ],
+            }
+        )
+        self.predictor._validate_tokenization(document=doc)
+        # missing token id
```
regan-huff, line 201: what's the scenario where this could happen?

regan-huff (approved):
This looks reasonable to me. I'm curious how you would characterize what's better about the SVM version vs. the original heuristic predictor.

I'm also wondering: to my eye, this still relies on an association of token id and word ids. I know the problem with the original predictor was this violated assertion: `assert None not in token_id_to_word_id.values()` Did we change something so that this error can't happen?

cmwilhelm:
I was branched off of this locally to work on timo-ifying. There seem to be some unspecified dependencies.

`pip install .[heuristic_predictors]` still leaves you with `joblib` uninstalled, not sure what else. I suspect that library is coming in transitively via some other `extra_requires` that we don't want to depend on.

cmwilhelm (changes requested):
May have found a showstopping bug. This seems to blow up on the very first paper I tried, see:

```python
>>> from mmda.parsers.pdfplumber_parser import PDFPlumberParser
>>> from mmda.predictors.heuristic_predictors.svm_word_predictor import SVMWordPredictor
>>> parser = PDFPlumberParser(split_at_punctuation=True)
>>> predictor = SVMWordPredictor.from_path("/tests/fixtures/svm_word_predictor/svm_word_predictor.tar.gz")
>>> doc = parser.parse("/output/no_bibs.pdf")
>>> predictor.predict(doc)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 225, in predict
    results = self.classifier.batch_predict(
  File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 107, in batch_predict
    all_features, word_id_to_feature_ids = self._get_features(words)
  File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 177, in _get_features
    dense_transformed = self.scaler.transform(dense_all)
  File "/usr/local/lib/python3.8/site-packages/sklearn/utils/_set_output.py", line 140, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/sklearn/preprocessing/_data.py", line 1233, in transform
    X = self._validate_data(
  File "/usr/local/lib/python3.8/site-packages/sklearn/base.py", line 565, in _validate_data
    X = check_array(X, input_name="X", **check_params)
  File "/usr/local/lib/python3.8/site-packages/sklearn/utils/validation.py", line 902, in check_array
    raise ValueError(
ValueError: Expected 2D array, got 1D array instead:
array=[].
Reshape your data either using array.reshape(-1, 1) if your data has a single feature or array.reshape(1, -1) if it contains a single sample.
>>>
```

The paper in question here is in this repo:
src/ai2_internal/bib_entry_detection_predictor/data/no_bibs.pdf


How many full documents was this predictor verified on?

[EDIT] If you dig a bit deeper with the above, it claims there are no candidate words with hyphens, but there are definitely hyphenated words in the pdf.

src/mmda/predictors/heuristic_predictors/svm_word_predictor.py
kyleclo: adding

src/mmda/predictors/heuristic_predictors/svm_word_predictor.py
```diff
@@ -0,0 +1,390 @@
...
+        # ws_token -> tokens
+        ws_token_id_to_tokens = defaultdict(list)
+        for token_id, ws_token_id in token_id_to_ws_token_id.items():
+            ws_token_id_to_tokens[ws_token_id].append(token_id)
+
+        # token -> all cluster tokens (inclusive)
+        # token_id_to_token_ids = {}
+        # for token_id, ws_token_id in token_id_to_ws_token_id.items():
+        #     candidate_token_ids = [i for i in ws_token_id_to_tokens[ws_token_id]]
+        #     token_id_to_token_ids[token_id] = candidate_token_ids
```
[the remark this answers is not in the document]
kyleclo: done, thx

src/mmda/predictors/heuristic_predictors/svm_word_predictor.py
```diff
@@ -0,0 +1,390 @@
...
+        for token_id in range(1, len(token_id_to_word_id)):
+            token = document.tokens[token_id]
+            word_id = token_id_to_word_id[token_id]
+            if word_id == current_word_id:
+                tokens_in_word.append(token)
+            else:
+                word = SpanGroup(
+                    spans=[span for token in tokens_in_word for span in token.spans],
+                    id=current_word_id,
+                    metadata=Metadata(text=word_id_to_text[current_word_id]),
```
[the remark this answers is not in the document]
kyleclo: ah great, thx for reminding, ill make this change

src/mmda/types/span.py
kyleclo: ah yes, this actually was broken. surprised it hadnt come up before. a minimal test case under the old code:

```
span = Span(0, 2)
Span.small_spans_to_big_span([span])
```
results in
```
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

whereas with the fix, you get:
```
Span(start=0, end=2, box=None)
```

tests/test_predictors/test_svm_word_predictor.py
kyleclo: someone reconstructs a document from scratch without an `id` field

one thought is -- this is starting to become a requirement for everything; maybe it should be baked into the `Document.from_json()` method that if `id` isnt specified, it gives a `warning` and automatically assigns an `id`?

what do u think

src/mmda/predictors/heuristic_predictors/svm_word_predictor.py
```diff
@@ -0,0 +1,390 @@
+"""
```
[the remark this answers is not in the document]
kyleclo: oh great point; i put it there since that's where the old word predictor was, but technically this should be something like an `sklearn predictor`. 

i can migrate to its own directory

kyleclo:
> I was branched off of this locally to work on timo-ifying. There seem to be some unspecified dependencies.
> 
> `pip install .[heuristic_predictors]` still leaves you with `joblib` uninstalled, not sure what else. I suspect that library is coming in transitively via some other `extra_requires` that we don't want to depend on.

thanks for catching, realized an issue w/ our GitHub CI is the build basically assumes installing everything. i made a new workflow that's specific to word predictor. I guess need to slowly build this up for all the other predictors



> May have found a showstopping bug. This seems to blow up on the very first paper I tried, see:
> 
> ```python
> >>> from mmda.parsers.pdfplumber_parser import PDFPlumberParser
> >>> from mmda.predictors.heuristic_predictors.svm_word_predictor import SVMWordPredictor
> >>> parser = PDFPlumberParser(split_at_punctuation=True)
> >>> predictor = SVMWordPredictor.from_path("/tests/fixtures/svm_word_predictor/svm_word_predictor.tar.gz")
> >>> doc = parser.parse("/output/no_bibs.pdf")
> >>> predictor.predict(doc)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 225, in predict
>     results = self.classifier.batch_predict(
>   File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 107, in batch_predict
>     all_features, word_id_to_feature_ids = self._get_features(words)
>   File "/usr/local/lib/python3.8/site-packages/mmda/predictors/heuristic_predictors/svm_word_predictor.py", line 177, in _get_features
>     dense_transformed = self.scaler.transform(dense_all)
>   File "/usr/local/lib/python3.8/site-packages/sklearn/utils/_set_output.py", line 140, in wrapped
>     data_to_wrap = f(self, X, *args, **kwargs)
>   File "/usr/local/lib/python3.8/site-packages/sklearn/preprocessing/_data.py", line 1233, in transform
>     X = self._validate_data(
>   File "/usr/local/lib/python3.8/site-packages/sklearn/base.py", line 565, in _validate_data
>     X = check_array(X, input_name="X", **check_params)
>   File "/usr/local/lib/python3.8/site-packages/sklearn/utils/validation.py", line 902, in check_array
>     raise ValueError(
> ValueError: Expected 2D array, got 1D array instead:
> array=[].
> Reshape your data either using array.reshape(-1, 1) if your data has a single feature or array.reshape(1, -1) if it contains a single sample.
> >>>
> ```
> 
> The paper in question here is in this repo: src/ai2_internal/bib_entry_detection_predictor/data/no_bibs.pdf
> 
> How many full documents was this predictor verified on?
> 
> [EDIT] If you dig a bit deeper with the above, it claims there are no candidate words with hyphens, but there are definitely hyphenated words in the pdf.

gah! nice catch, sry I only tested on two column papers. one-column papers rarely have hyphens that need correcting (the test example doesn't have one). I added a handling for this + test.

tests/test_predictors/test_svm_word_predictor.py
```diff
@@ -0,0 +1,286 @@
...
+            else:
+                # otherwise, most token units should be substring of word
+                overlap = np.mean([token.text in word.text for token in tokens_in_word])
+                self.assertGreaterEqual(overlap, 0.5)
+                # spans should match up though
+                self.assertEqual(
+                    [span for token in tokens_in_word for span in token.spans],
+                    word.spans,
+                )
+        # uncomment for debugging:
```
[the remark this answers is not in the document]
kyleclo: thankss

tests/test_predictors/test_svm_word_predictor.py
```diff
@@ -0,0 +1,320 @@
...
+        with open(
+            os.path.join(
+                self.fixture_path,
+                "types/20fdafb68d0e69d193527a9a1cbe64e7e69a3798__pdfplumber_doc.json",
+            )
+        ) as f_in:
+            doc_dict = json.load(f_in)
+            self.doc = Document.from_json(doc_dict=doc_dict)
+
+    def test_predict_no_hyphen(self):
```
kyleclo, line 155: @cmwilhelm added this test to handle failure case u found

cmwilhelm: tyty

cmwilhelm (approved):
Seems to work now!

tests/test_predictors/test_svm_word_predictor.py
kyleclo: > I'm also wondering: to my eye, this still relies on an association of token id and word ids. I know the problem with the original predictor was this violated assertion: `assert None not in token_id_to_word_id.values()` Did we change something so that this error can't happen?

great catch. we changed 2 things:
* different implementation. the previous approach required this iterative "building up" of the `token_id_to_word_id` map, and this assertion was necessary to make sure we didn't miss writing rules for certain cases when building this up.  the current implementation takes a different approach to this incremental build-up. it automatically assumes all tokens are words themselves, so it seeds `token_id_to_word_id` with non-null values from the get-go. 
* different ideology. i realized visiting this problem properly that we should absolutely allow for tokens to be skipped, if necessary, which would mean there would be tokens mapped to no word. this implementation doesn't do this, but conceptually, we need to allow for tokens mapped to nothing if we want to future handle cases like hallucinated/fake tokens in PDFs

PR #248, opened 2023-05-17
TIMO-ifies heuristic DWP predictor

cmwilhelm:
Same behavior as implementation in SPP, but
can be stood up as a TIMO service.


cmwilhelm:
fyi @kyleclo @regan-huff

PR #249, opened 2023-05-30
TIMO-ifies pysbd sentences predictor

cmwilhelm:


src/ai2_internal/pysbd_sentences/integration_test.py
```diff
@@ -0,0 +1,51 @@
...
+    doc_file = resolve("test_doc.json")
+    with open(doc_file) as f:
+        doc = Document.from_json(json.load(f))
+
+    tokens = [api.SpanGroup.from_mmda(sg) for sg in doc.tokens]
+    pages = [api.SpanGroup.from_mmda(sg) for sg in doc.pages]
+
+    return Instance(
+        symbols=doc.symbols,
+        tokens=tokens,
```
soldni, line 37: @cmwilhelm in my experience, pysbd works best off words rather than tokens. we can do an initial deployment for now, and then add  words once @kyleclo is done with #247.

soldni (approved):

src/ai2_internal/pysbd_sentences/integration_test.py
cmwilhelm: Ah, maybe I misread the code in timo_scim, but I thought you were sending it tokens over there?

cmwilhelm: Ah shoot I see it now. You have your own word predictor, and the pysbd predictor conditionally uses words when available. Well shoot.

cmwilhelm: My understanding is Kyle's words predictor (and the existing on in mmda) don't do anything except figure out whether hyphenated words are line breaks or compound words. No text cleanup etc. So yours is already probably much more sophisticated, probably not appropriate to sub in Kyle's for this use case.

soldni: mine is essentially similar to kyle, only additional workaround is merging small-caps. but i'll sync with kyle to see what's missing

PR #251, opened 2023-06-02
add new linker and bump mmda version

dmh43:
Once i deploy then new linker to spp I can run the docs through the pipeline again and see what changed in the eval.

### DIFF
--- a/src/ai2_internal/config.yaml
+++ b/src/ai2_internal/config.yaml
@@ -191,7 +191,7 @@ model_variants:
     extras_require: ["dev", "citation_links"]
 
     # Full S3 path to tar.gz'ed artifacts archive
-    artifacts_s3_path: s3://ai2-s2-analytics/danyh/extract-stuff-from-pdfs/artifacts/links/v1.tar.gz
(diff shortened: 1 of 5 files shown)


src/ai2_internal/config.yaml
```diff
@@ -191,7 +191,7 @@ model_variants:
     extras_require: ["dev", "citation_links"]
 
     # Full S3 path to tar.gz'ed artifacts archive
-    artifacts_s3_path: s3://ai2-timo-registry/model-artifacts/citationlinks/v0.tar.gz
+    artifacts_s3_path: s3://ai2-s2-analytics/danyh/extract-stuff-from-pdfs/artifacts/links/v1.tar.gz
```
regan-huff, line 194: maybe go ahead and put a copy of this in the timo model-artifacts directory and use that s3 path here? since you've distinguished it as v1 I think that won't be confusing and you don't have to wipe out the v0 artifact

regan-huff (commented):
✨

src/ai2_internal/config.yaml
dmh43: good call! done!

geli-gel (approved):
nice use of PredictorConfig 👍🏽

PR #254, opened 2023-06-08
Kylel/stringify

kyleclo:


src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+    document: Document,
+    use_word_text: Optional[bool] = True,
+    replace_newlines_with: Optional[str] = " ",
+    join_words_with: Optional[str] = " ",
+    normalize_whitespace: Optional[bool] = True,
+    allow_partial_word_match: Optional[bool] = True,
+    allow_disjoint_spans: Optional[bool] = True,
+    include_symbols_between_disjoint_spans: Optional[bool] = False,
+) -> str:
+    """Requires doc.words to exist. This technique applies 4 main steps:
```
regan-huff, line 33: where does `doc.words` come from?

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+                f"span_group {span_group} has disjoint spans but allow_disjoint_spans is False"
+            )
+        else:
+            logger.warning(f"span_group {span_group} has disjoint spans")
+            if include_symbols_between_disjoint_spans:
+                # TODO: easiest is probably to convert disjoint spans into a single longer span
+                raise NotImplementedError
+
+    # if matched words are disjoint, what should we do with the in-between symbols when stringifying?
+    if Span.are_disjoint(spans=[span for word in matched_words for span in word.spans]):
```
regan-huff, line 84: can you help me understand how this situation would arise? We were given SpanGroup spans that were overlapping, we fetched words associated with those spans, we're now checking the spans on the associated words, and they're disjoint? Which spans do we care about? Maybe that's what you mean with your comment on L97

src/mmda/utils/stringify.py
regan-huff: I may not have perfectly followed the conversation in the earlier SVM ticket/slack, but what is happening here as far as punctuation? We expect that infelicitous hyphens have been removed already in the words? What about the issue as far as, say, author names having undesirable commas clumped with them? How is that solved?

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+            if include_symbols_between_disjoint_spans:
+                # TODO: easiest is probably to convert disjoint spans into a single longer span
+                raise NotImplementedError
+
+    # TODO: actually, maybe refactor this. it doesnt matter if query spangroup is disjoint
+    #       the actual handling should just happen if it's disjoint words.
+
+    # define util function that produces text from a word
+    _stringify_word = lambda word: word.text if use_word_text else None
+    if use_word_text is False:
```
regan-huff, line 102: what text would a caller want used if not word text?

src/mmda/utils/stringify.py
regan-huff: I'm seeing some tests relevant to this, where the symbols include hyphens and line breaks that aren't in the words, and where serial commas end up as their own word. Is that the main resolution of the earlier question? The word predictor doesn't attempt to combine tokens with no whitespace between, but it does (we hope) successfully remove hyphens related to columns and pagination? And then it is up to the caller of stringify to have accurate enough span group boundaries to not get stray commas if they aren't wanted?

regan-huff: Is there a way we can see if that actually works? I have trouble with this concept that we might be using the spans that come with the words to determine the stringify result, and that these are different from the spans the caller actually included in request

src/mmda/types/span.py
```diff
@@ -53,3 +53,14 @@ def small_spans_to_big_span(cls, spans: List["Span"]) -> "Span":
             end=end,
             box=Box.small_boxes_to_big_box(boxes=[span.box for span in spans]),
         )
+
+    def is_overlap(self, other: "Span") -> bool:
+        return self.start < other.end and self.end > other.start
```
geli-gel, line 58: what about if `self.start > other.end and self.end > other.start` (other encompasses self)? would it make sense to put `is_overlap_1d` ([this](https://github.com/allenai/mmda/blob/58a2bfe40c8078525bc616c2665e9d5a9ae3834d/src/mmda/types/box.py#L14)) in a more common place and use it here same as `box`? Is there something different about trying to check overlap that way when dealing with Spans?

cmwilhelm (changes requested):
Some comments and questions.

src/mmda/utils/stringify.py
kyleclo: L84 is concerned w/ the case where, let's say an input query `SpanGroup` looks like:
```
[(0, 10), (15, 20)]
```
in this case, let's say it overlaps with the following words:
```
word1 -> (0, 10)
word2 -> (11, 14)
word3 -> (15, 20)
```
should the `stringify()` return the text from `word2` or just the text from `word1` and `word3`? If the latter, how should it stitch together?

I'm not sure when this would occur in SPP; it's sthn that @soldni  & I realized while implementing this that the intended behavior of how to handle `word2` isn't clear, so making it explicit here.

src/mmda/utils/stringify.py
kyleclo: chatting w/ @soldni , there are cases (e.g. debugging) where one would want to actually defer to the original `doc.symbols [ start : end ]` string. this wasn't a critical use-case so didn't implement this & left it behind `NotImplementedError`, but wanted to call this out in implementation that it's a possible configurable choice (though default should be `word.text`)

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+def stringify_span_group(
+    span_group: SpanGroup,
+    document: Document,
+    use_word_text: Optional[bool] = True,
+    replace_newlines_with: Optional[str] = " ",
+    join_words_with: Optional[str] = " ",
+    normalize_whitespace: Optional[bool] = True,
+    allow_partial_word_match: Optional[bool] = True,
+    allow_disjoint_spans: Optional[bool] = True,
+    include_symbols_between_disjoint_spans: Optional[bool] = False,
```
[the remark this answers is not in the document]
kyleclo: ohh TIL!  i thought `Optional` was the typing norm for these cases, fixed

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+        [word.start == span_group.start for word in matched_words]
+    )
+    match_any_word_end = any([word.end == span_group.end for word in matched_words])
+    if allow_partial_word_match:
+        if not match_any_word_start:
+            logger.warning(
+                f"span_group {span_group}'s start index doesnt match w/ start of any word. output string may thus include more text than expected."
+            )
+        if not match_any_word_end:
+            logger.warning(
```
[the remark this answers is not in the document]
kyleclo: sounds good; fixed

tests/test_utils/test_stringify.py
```diff
@@ -0,0 +1,327 @@
...
+                    {"start": 5, "end": 17},
+                ]
+            }
+        )
+        self.assertEqual(
+            stringify_span_group(span_group=query_span_group, document=doc),
+            "is a test.",
+        )
+
+    def test_how_words_relate_to_stringify(self):
```
kyleclo, line 219: @regan-huff @cmwilhelm does this test help explain the confusing behavior about `words` and `stringify()`?

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+    # if query has disjoint spans, what should we do with the in-between symbols when stringifying?
+    if Span.are_disjoint(spans=span_group.spans):
+        if not allow_disjoint_spans:
+            raise ValueError(
+                f"span_group {span_group} has disjoint spans but allow_disjoint_spans is False"
+            )
+        else:
+            logger.warning(f"span_group {span_group} has disjoint spans")
```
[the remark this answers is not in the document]
kyleclo: done

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+                raise NotImplementedError
+
+    # if matched words are disjoint, what should we do with the in-between symbols when stringifying?
+    if Span.are_disjoint(spans=[span for word in matched_words for span in word.spans]):
+        if not allow_disjoint_spans:
+            raise ValueError(
+                f"span_group {span_group} intersects words {matched_words} which have disjoint spans but allow_disjoint_spans is False"
+            )
+        else:
+            logger.warning(
```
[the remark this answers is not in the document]
kyleclo: done

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+            f"Havnt figured out how to do this yet. Only supports use_word_text=True for now."
+        )
+
+    # define util function that replaces all whitespace with a single character
+    _normalize_whitespace = lambda text: " ".join(text.split())
+
+    # stringify!
+    prev = matched_words[0]
+    text_from_words_including_whitespace: List[str] = [_stringify_word(word=prev)]
+    for i in range(1, len(matched_words)):
```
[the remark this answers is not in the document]
kyleclo: your'e right; fixed

src/mmda/types/span.py
```diff
@@ -53,3 +53,16 @@ def small_spans_to_big_span(cls, spans: List["Span"]) -> "Span":
             end=end,
             box=Box.small_boxes_to_big_box(boxes=[span.box for span in spans]),
         )
+
+    def is_overlap(self, other: "Span") -> bool:
+        is_self_before_other = self.start < other.end and self.end > other.start
+        is_other_before_self = other.start < self.end and other.end > self.start
```
cmwilhelm, line 59: You know, I think I misread this the first time and your original condition was sufficient, i.e.

`self.start < other.end and self.end > other.start`

cmwilhelm (approved):
I rescind one of my prior comments but this looks good.

src/mmda/utils/stringify.py
```diff
@@ -0,0 +1,128 @@
...
+        text_from_words_including_whitespace.append(_stringify_word(word=current))
+        prev = current
+
+    # final step of formatting the text to return
+    candidate_text = "".join(text_from_words_including_whitespace)
+    if normalize_whitespace:
+        candidate_text = _normalize_whitespace(candidate_text)
+    else:
+        if replace_newlines_with is not None:
+            candidate_text = candidate_text.replace("\n", replace_newlines_with)
```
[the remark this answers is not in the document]
kyleclo: i've refactored so these are two independent calls; the newline replacement happens first. there's still interaction which is intentional. i've added a test case that clarifies this; pls see if that's ok thx

PR #255, opened 2023-06-09
Grobid bib refs

geli-gel:
Adding in Grobid citation-mentions and target info to output box_group metadata, also added grobid ids to bib entries -
first step to getting these and relations coming from SPP-Grobid API (https://github.com/allenai/scholar/issues/36975)


regan-huff (approved):
This looks good to me!

PR #259, opened 2023-07-05
Kylel/2023/spangroup overlap bug

kyleclo:


src/mmda/types/annotation.py
```diff
@@ -77,40 +75,44 @@ def __getattr__(self, field: str) -> List["Annotation"]:
...
-            boxes: List[Box],
-            id: Optional[int] = None,
-            doc: Optional['Document'] = None,
-            metadata: Optional[Metadata] = None,
+        self,
+        boxes: List[Box],
+        id: Optional[int] = None,
+        doc: Optional["Document"] = None,
+        metadata: Optional[Metadata] = None,
+        allow_overlap: Optional[bool] = False,
```
cmwilhelm, line 85: `allow_overlap=False` as the default is a behavioral change and will almost certainly break stuff in practice. Can we switch the default to `True`?

src/mmda/types/annotation.py
```diff
@@ -150,25 +152,30 @@ def type(self, type: Union[str, None]) -> None:
...
-            id: Optional[int] = None,
-            doc: Optional['Document'] = None,
-            metadata: Optional[Metadata] = None,
+        self,
+        spans: List[Span],
+        box_group: Optional[BoxGroup] = None,
+        id: Optional[int] = None,
+        doc: Optional["Document"] = None,
+        metadata: Optional[Metadata] = None,
+        allow_overlap: Optional[bool] = False,
```
cmwilhelm, line 162: Same story here as `BoxGroup` -- `False` as the default is a breaking change in behavior, and there's no telling how much data we already have that violates it.

src/mmda/types/box.py
```diff
@@ -95,11 +116,6 @@ def coordinates(self) -> Tuple[float, float, float, float]:
     def center(self) -> Tuple[float, float]:
         return self.l + self.w / 2, self.t + self.h / 2
 
-    @property
-    def xywh(self) -> Tuple[float, float, float, float]:
-        """Return a tuple of the (left, top, width, height) format."""
-        return self.l, self.t, self.w, self.h
```
cmwilhelm, line 101: Let's leave the method in but add a logging.warning() with a deprecation notice. This may break clients of mmda that use `.xywh`

src/mmda/types/box.py
```diff
@@ -120,18 +136,109 @@ def get_absolute(self, page_width: int, page_height: int) -> "Box":
...
+    def is_overlap(
+        self, other: "Box", x: float = 0.0, y: float = 0, center: bool = False
+    ) -> bool:
         """
         Whether self overlaps with the other Box object.
         x, y distances for padding
         center (bool) if True, only consider overlapping if this box's center is contained by other
         """
+        if self.page != other.page:
+            return False
```
cmwilhelm, line 148: 🤯

src/mmda/types/indexers.py
```diff
@@ -68,23 +69,36 @@ def _ensure_disjoint(self) -> None:
...
+                        span.start, span.end
+                    )
+                ]
+                if len(match_ids) > 1:
+                    matches = [self._sgs[match_id].to_json() for match_id in match_ids]
                     raise ValueError(
-                        f"Detected overlap with existing SpanGroup(s) {matches} for {span_group}"
+                        f"Detected overlap! While processing the Span {span} as part of query SpanGroup {span_group.to_json()}, we found that it overlaps with existing SpanGroup(s):\n"
+                        + "\n".join(
+                            [f"\t{i}\t{m} " for i, m in zip(match_ids, matches)]
```
cmwilhelm, line 83: Thanks this is much more informative output.

cmwilhelm (changes requested):
Changes generally make sense to me; I called out the places I believe to be breaking and have suggested alternatives.

Whenever this is good to go I also suggest we merge/rebase in the latest `main` (to get CI to rerun) and bump the MMDA version in pyproject.toml

regan-huff (approved):
I would love for this PR to include or be accompanied by documentation for the current best expectations for how SpanGroups and BoxGroups will be used.

I also support the more conservative approach Chris suggests of leaving the defaults in a state that will allow existing/old data to be loaded from the annotation store without throwing an error. But perhaps the documentation can explain that the defaults are for backwards compatibility.

I think what we'll try to do to handle the possible impact of these model changes is attempt to update the mmda version for the whole Semantic Reader more or less simultaneously, so we can hopefully have a tight response loop if we do discover some data incompatibilities with the updated mmda models.

PR #261, opened 2023-07-06
adds a timo wrapper for the svm word predictor

cmwilhelm:


src/ai2_internal/svm_word_predictor/integration_test.py
```diff
@@ -0,0 +1,81 @@
...
+        doc_file = resolve("test_doc.json")
+        with open(doc_file) as f:
+            doc = Document.from_json(json.load(f))
+
+        tokens = [api.SpanGroup.from_mmda(sg) for sg in doc.tokens]
+        instances = [Instance(symbols=doc.symbols, tokens=tokens)]
+        predictions = container.predict_batch(instances)
+        prediction = predictions[0]
+
+        self.assertTrue(len(prediction.words) <= len(tokens))
```
regan-huff, line 81: is there anything else we can or should assert about `prediction.words`?

regan-huff (approved):
What a nice wrapper!

src/ai2_internal/svm_word_predictor/integration_test.py
cmwilhelm: /shrug

PR #264, opened 2023-07-11
Kylel/fix index error word pred

kyleclo:
the error here is in the word predictor's `get_features()`. 

at a high-level, the idea of the word predictor is that it loops over each hyphen in a candidate word and predicts whether it should be kept or removed. So a word like `wizard-of-oz` would trigger two predictions, one for each hyphen. The way the model works is it looks at a window surrounding that word to derive features that help with prediction. So for `wizard-of-oz`, the first hyphen's featurization is `wizard, of` and the second hyphen's featurization is `of, oz`.

the problem we're seeing is one due to tokenization. some PDFs result in weird tokenizations where the `token.text.strip() == ""`. I took a look into this, and it's not a bug introduced in our `PDFPlumberTokenizer`, but an upstream issue with just... pdfs in general I guess. When these empty tokens are introduced, it causes situations in which you have words that end with hyphenation. For example, what if there were an empty token between `wizard-of-` and `oz`. 

there are several ways to fix this. one is to fix tokenization, but we'd have to re-drive PDFs, which is too costly at the moment. another is to refactor & retrain the model to know how to properly handle even empty tokenization cases. this is also too costly right now.

so what I've done here is a bit hacky but it's essentially -- add a cleaning step at the beginning of the word predictor's `.predict()` in which we create a new Document that we operate over. as we catch weird issues with tokenization in the future, we can add to this function. all word prediction happens on top of this new document. at the end, we return words as basically start-end spans, so the removal of tokens shouldn't affect how we perform `document.annotate(words=words)` on the original, uncleaned document.

### DIFF
--- a/src/mmda/predictors/sklearn_predictors/svm_word_predictor.py
+++ b/src/mmda/predictors/sklearn_predictors/svm_word_predictor.py
@@ -9,6 +9,7 @@ classifier to predict whether hyphenated segments should be considered a single 
 
 """
 
+import logging
 import os
 import re
 import tarfile
@@ -30,6 +31,8 @@ from mmda.predictors.sklearn_predictors.base_sklearn_predictor import (
     BaseSklearnPredictor,
 )
 from mmda.types import Document, Metadata, Span, SpanGroup
+
+logger = logging.getLogger(__name__)
 
 
 class IsWordResult:
@@ -117,6 +120,11 @@ class SVMClassifier:
         return classifier
 
     def batch_predict(self, words: List[str], threshold: float) -> List[IsWordResult]:
+        if not all(
+            [not word.startswith("-") and not word.endswith("-") for word in words]
+        ):
+            raise ValueError("Words should not start or end with hyphens.")
+
         all_features, word_id_to_feature_ids = self._get_features(words)
         all_scores = self.estimator.decision_function(all_features)
         results = []
@@ -220,6 +228,9 @@ class SVMWordPredictor(BaseSklearnPredictor):
         return predictor
 
     def predict(self, document: Document) -> List[SpanGroup]:
+        # clean input
+        document = self._make_clean_document(document=document)
+
         # validate input
         self._validate_tokenization(document=document)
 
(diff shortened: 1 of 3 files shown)


src/mmda/predictors/sklearn_predictors/svm_word_predictor.py
```diff
@@ -475,7 +526,12 @@ def _create_words(
         new_word_id = 0
         for token_id in range(1, len(token_id_to_word_id)):
             token = document.tokens[token_id]
-            word_id = token_id_to_word_id[token_id]
+            word_id = token_id_to_word_id.get(token_id)
+            if word_id is None:
```
kyleclo, line 530: not sure if this warning makes the logs too noisy; was thinking it's useful to know which PDFs need to be re-driven in future when we fix tokenizer

geli-gel: Hmm, shouldn't this loop be `for token_id in range(1,len(document.tokens))` to go through all tokens including those that might not have a `word_id`?

geli-gel (approved):
word

src/mmda/predictors/sklearn_predictors/svm_word_predictor.py
cmwilhelm: I'd change this to `logger.debug()`. This will indeed be quite noisy.

cmwilhelm (approved):
LGTM.

Will this impact the ability of `utils.stringify()` to make strings from words when the words are annotated onto the OG document?

cmwilhelm (approved):
Should bump mmda version +0.0.1 with this, as a bug fix.

kyleclo:
> LGTM.
> 
> Will this impact the ability of `utils.stringify()` to make strings from words when the words are annotated onto the OG document?

it won't. the new document is an object internal to the word predictor, and the key is that it reuses the original document's `.symbols`. since all things are stored as span-groups, which basically points to start-end positions, as long as `doc.symbols` remains the same, then there's no difference between words based on spans on the new doc vs old doc.

an alternative implementation would've been to modify the existing doc with something like a hidden field, e.g. `doc.annotate(_new_tokens)` and perform the operations wrt to those to derive `words`, then do `doc.remove(_new_tokens)` as a final step.

i think there's no difference between the two approaches, but keeping things separate was a bit easier for debugging

PR #267, opened 2023-07-13
constrain vram utilization in citation mentions

cmwilhelm:
Unbounded input sizes (within a page) were 
being sent to the GPU for inference. This changeset
forces a constrained batch size of 512 tokens.

Context: the citation mentions TIMO endpoint
has been suffering since pointing SPP prod at
it. Broadly it shows very high vram utilization
that is not solved by reducing worker count
per machine, and very high inference times. Both
these things were seen as well in the bibentry detector,
resolved with a similar patch.

Unlike that model, however, I was not able to
reproduce memory spikes in a test setting with
real papers, likely due to the relative rarity of papers
with excessive numbers of tokens on individual pages.
The issue was reproducible, however, by fabricating
some pathological data and sending that to test
servers. Seen below, the first spikes in activity are
real papers sampled from SPP's recent work queue,
and the latter spikes in activity are the same papers
rewritten as being "one" page long, with all tokens.

Unbounded, existing solution, 1 worker:
Screen Shot 2023-07-13 at 4 23 27 PM

With this PR's changes, 1 worker:
Screen Shot 2023-07-13 at 4 23 47 PM

### DIFF
--- a/src/mmda/predictors/hf_predictors/mention_predictor.py
+++ b/src/mmda/predictors/hf_predictors/mention_predictor.py
@@ -5,7 +5,7 @@ from typing import Dict, Iterator, List, Optional
 
 from optimum.onnxruntime import ORTModelForTokenClassification
 import torch
-from transformers import AutoModelForTokenClassification, AutoTokenizer
+from transformers import AutoModelForTokenClassification, AutoTokenizer, BatchEncoding
 
 from mmda.types.annotation import Annotation, SpanGroup
 from mmda.types.document import Document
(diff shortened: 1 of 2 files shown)


src/mmda/predictors/hf_predictors/mention_predictor.py
```diff
@@ -91,12 +91,24 @@ def predict_page(self, page: Annotation, counter: Iterator[int], print_warnings:
...
-        del inputs["overflow_to_sample_mapping"]
+        )
 
         with torch.no_grad():
-            outputs = self.model(**inputs)
-        prediction_label_ids = torch.argmax(outputs.logits, dim=-1)
+            # Control device memory use to predictable levels
+            # by limiting size of batches sent to it.
+            prediction_label_ids = []
+            for index, sequence in enumerate(inputs["input_ids"]):
```
dmh43, line 100: Why not use a larger batch size here?

dmh43 (approved):
Looks right to me

src/mmda/predictors/hf_predictors/mention_predictor.py
cmwilhelm: The small batch size had no impact on inference speed, but reduces the max footprint of any one copy of the model. This means we can jam more copies onto one device, and shoudl be able to reap greater concurrency as a result.

PR #270, opened 2023-07-20
Bib detector fixes

cmwilhelm:
Fixes bibentry detector issue with orphaned boxes

RE: https://github.com/allenai/scholar/issues/35308

    Sometimes the detector predicts boxes that don't intersect
    any tokens on the page. In these cases we should discard
    the box, as it doesn't map to text and causes a runtime error
    otherwise.


geli-gel (approved):
LGTM!

PR #271, opened 2023-07-27
GrobidAugmentExistingDocumentParser: raise errors for bad Grobid client statuses

geli-gel:
Part of addressing https://github.com/allenai/scholar/issues/37582
partner PR to https://github.com/allenai/spp-grobid-api/pull/29
todo: 
- [ ] tag 0.9.8 mmda version

### DIFF
--- a/src/mmda/parsers/grobid_augment_existing_document_parser.py
+++ b/src/mmda/parsers/grobid_augment_existing_document_parser.py
@@ -44,7 +44,7 @@ class GrobidAugmentExistingDocumentParser(Parser):
         for field in REQUIRED_DOCUMENT_FIELDS:
             assert field in doc.fields
 
-        (_, _, xml) = self.client.process_pdf(
+        (_, status, xml) = self.client.process_pdf(
             "processFulltextDocument",
             input_pdf_path,
             generateIDs=False,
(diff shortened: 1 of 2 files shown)


src/mmda/parsers/grobid_augment_existing_document_parser.py
```diff
@@ -55,15 +55,19 @@ def parse(self, input_pdf_path: str, doc: Document, xml_out_dir: Optional[str] =
...
-        if xml_out_dir:
-            os.makedirs(xml_out_dir, exist_ok=True)
-            xmlfile = os.path.join(
-                xml_out_dir, os.path.basename(input_pdf_path).replace(".pdf", ".xml")
-            )
-            with open(xmlfile, "w") as f_out:
-                f_out.write(xml)
-
-        self._parse_xml_onto_doc(xml, doc)
+        if status == 200:
```
regan-huff, line 58: Something feels a bit weird about the division of responsibilities here to me, and I'm wondering if this added complexity of error raising is just starting to show the weirdness a bit more.

In my mind, the mmda parser should just accept grobid xml and return MMDA document. It seems off that you have to make these client error changes here -- they aren't really related to parsing.

It makes more sense to me that something in the spp-grobid-api layer would actually run the grobid client and handle things like status code checking and error throwing, and then would just call this function if it had a grobid result. I don't think it's necessary to try to make this change right now with so many other issues in play, and I don't remember why this was structured this way in the first place -- I just wanted to take the opportunity to call out that it seems awkward now and maybe we should change it if we see an opportunity.

regan-huff (approved):

src/mmda/parsers/grobid_augment_existing_document_parser.py
geli-gel: True. It feels awkward, and I didn't feel very good making this PR, but I think for now it's helpful

PR #273, opened 2023-08-07
merge boxes from citation mention model

dmh43:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.9.9'
+version = '0.10.0'
```
geli-gel, line 3: I swear I'm always confused by semver so i might be wrong again but shouldn't this be 0.9.10 (only increase the last number because it is a minor change that doesn't really affect the whole MMDA library as a whole?)

src/ai2_internal/citation_mentions/interface.py
```diff
@@ -39,6 +39,34 @@ class PredictorConfig(BaseSettings):
...
+    w = max(b.l + b.w for b in boxes) - l
+    h = max(b.t + b.h for b in boxes) - t
+    return mmda.types.box.Box(l=l, t=t, w=w, h=h, page=boxes[0].page)
+
+def merge_boxes(boxes):
+    boxes_by_line = group_by_line(boxes)
+    return [calc_bounding_box(line_boxes) for line_boxes in boxes_by_line]
+
+def merge_boxes_of_sg(sg):
+    sg.box_group.boxes = merge_boxes(sg.box_group.boxes)
```
geli-gel, line 58: why does this work 🤣 I thought that the SpanGroups we're dealing with here have box info only on the Span, so this line would have to be sg.box_group.boxes = merge_boxes([box for box in each sg.spans]) and then delete the boxes from the Spans... but I've seen your images and that the boxes are being merged so... Can you just confirm that the final output only has boxes on sg.box_group and no boxes on sg.spans[0].box? otherwise we'll get an error in SPP

geli-gel: ohhh maybe it's because the tokens are coming from the annotation store where we convert them into SpanGroupWithBoxGroup format?? I'm assuming that's it, and the tokens we start out with here have box info only on BoxGroup. but yeah maybe just confirming we only have box info in that one spot would be a good thing to check...

src/ai2_internal/citation_mentions/interface.py
```diff
@@ -39,6 +39,34 @@ class PredictorConfig(BaseSettings):
     pass
 
 
+def group_by_line(boxes):
+    boxes = sorted(boxes, key=lambda box: box.t)
+    return [list(line_boxes) for t, line_boxes in groupby(boxes, key=lambda box: box.t)]
```
geli-gel, line 44: Are you seeing that the line grouping is working well? I think it could be better if we had `rows` passed into this model, so that we can take advantage of pdfplumber's grouping by lines which I think is more robust than requiring exact `t` coordinates matching up (see https://github.com/allenai/mmda/blob/5c2fb76a3ec017f20f27dca056ce57a425961296/src/mmda/parsers/pdfplumber_parser.py#L370) -- or maybe just box.t +/- a little amount calculated from getting average token heights? (see https://github.com/allenai/mmda/blob/6ab9004e0df5466201396163da8be6c34e222d2d/src/mmda/predictors/heuristic_predictors/figure_table_predictors.py#L53) -- just options, but if you're seeing good results with this grouping I wouldn't worry about making this more complicated

pyproject.toml
dmh43: i guess I considered it a breaking change, but I think you're right since no one actually depends on the specific behavior that's changing in this case.

src/ai2_internal/citation_mentions/interface.py
dmh43: it actually seems to work well! if the lines dont have the same `t` then I think it's safest to just not do anything.

pyproject.toml
geli-gel: yeah, I think we usually consider the ai2 model updates as minor

geli-gel (approved):

src/ai2_internal/citation_mentions/interface.py
```diff
@@ -39,6 +41,35 @@ class PredictorConfig(BaseSettings):
...
+    h = max(b.t + b.h for b in boxes) - t
+    return Box(l=l, t=t, w=w, h=h, page=boxes[0].page)
+
+def merge_boxes(boxes):
+    boxes_by_line = group_by_line(boxes)
+    return [calc_bounding_box(line_boxes) for line_boxes in boxes_by_line]
+
+def merge_boxes_of_sg(sg):
+    boxes = merge_boxes([span.box for span in sg.spans])
+    sg.box_group = api.BoxGroup(boxes=[api.Box.from_mmda(box) for box in boxes]).to_mmda()
```
geli-gel, line 61: This is what I initially expected it would need to look like, but now, won't your local run not work, even if tt verify does? I think it is a good idea to check what exactly SPP passes to the citation_mention model by running something locally. Because I am so unsure of what happens there, and feel like it's possible we need to keep this the way you first had it, and instead update the fixture.

geli-gel: Actually I take that back -- we should be able to trust the config file in SPP which says that we are sending tokens with Spans on Boxes: https://github.com/allenai/spp/blob/3a394aa90265f2a852708abe1038f5993d3760e7/src/main/resources/spp/base.conf#L229

geli-gel (changes requested):
Unfortunately we have to make a decision and change something here... i think we have 2 options:
1) change SPP config to provide as input to citation_mentions `tokens` that have `SpanGroupWithBoxGroup` format, and revert your latest commit, needing to then update the fixture to match, or
2) add another step in `merge_boxes_of_sg` to delete the boxes that with the latest commit are still in Span.box (because having boxes in both places will not be allowed by SPP)

src/ai2_internal/citation_mentions/interface.py
```diff
@@ -39,6 +41,33 @@ class PredictorConfig(BaseSettings):
...
+def merge_boxes_of_sg(sg):
+    sg.box_group.boxes = merge_boxes(sg.box_group.boxes)
+
+def all_spans_close(sg):
+    spans = sorted(sg.spans, key=lambda span: span.start)
+    return all(span.end <= next_span.start <= span.end + 5 for span, next_span in zip(spans, spans[1:]))
+
+def build_box_group(sg):
+    boxes = [span.box for span in sg.spans]
+    return api.BoxGroup(boxes=[api.Box.from_mmda(box) for box in boxes]).to_mmda()
```
geli-gel, line 68: is it necessary to convert .from_mmda .to_mmda?

geli-gel (approved):

PR #274, opened 2023-08-14
group boxes into lines with tolerance

dmh43:
This PR groups boxes into lines without assuming a perfect match of `box.t`. We use the 3rd decimal point which seems small enough but also big enough to catch most cases.

This PR also merges adjacent boxes belonging to a mention which might have 2 spans that are far apart. Instead of not doing anything in that case, it only merges boxes with associated spans that are close.


geli-gel (approved):

geli-gel:
Looks like this will improve things but wondering if we could take advantage of PDFPlumber's line segmentation (`rows`) downstream to decide how to draw boxes

geli-gel:
Just realized a version bump is needed here, I'm setting it to 0.9.11 in my PR but idk when mine will merge so feel free to take 0.9.11 if I don't get to it first

PR #276, opened 2023-08-29
Fix for sg overlap error in box_groups_to_span_groups when center=true

geli-gel:
Solution for https://github.com/allenai/scholar/issues/38452
The problem was in running `box_groups_to_span_groups` with `center=True`. We want to keep `center=True` because it allows us to get good texts in most cases when converting from Grobid box groups to MMDA spangroups. However in some cases it was causing a problem.
The overlapping spangroups error seen from spp-grobid was always a single character span found in 2 different spangroups. This was due to single character tokens that were sometimes not found to be overlapping with the box when running `allocate_overlapping_tokens_for_box`, but then got swallowed up with MergeSpans, and we weren't accounting for this. We now update the dictionary containing the remaining allocatable tokens _after_ `merge_neighbor_spans_by_symbol_distance` is used to derive a SpanGroup with all the tokens used (instead of just the tokens whose centers overlapped with the original box)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.9.12'
+version = '0.9.13'
 description = 'MMDA - multimodal document analysis'
 authors = [
     {name = 'Allen Institute for Artificial Intelligence', email = 'contact@allenai.org'},
--- a/src/mmda/utils/tools.py
+++ b/src/mmda/utils/tools.py
@@ -121,8 +121,7 @@ def box_groups_to_span_groups(
         derived_spans = merge_spans.merge_neighbor_spans_by_symbol_distance()
 
         # tokens overlapping with derived spans:
-        doc.annotate(derived_span_group=[SpanGroup(spans=derived_spans)])
-        sg_tokens = doc.derived_span_group[0].tokens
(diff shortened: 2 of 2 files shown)


geli-gel:
To reproduce the overlap error download grobid xml [here](https://gist.github.com/geli-gel/9e41b0c0b276eecd774094e2aa8d62dc) and run the following (this runs in a nb located in examples/grobid_augment_existing_document_parser/notebook.ipynb):
```py
PDF_PATH = 'pdfs/3d5c8c04f42be8bc2fd6038e6f4099bbcfaa0c54.pdf'
XML_PATH = '3d5c8c04f42be8bc2fd6038e6f4099bbcfaa0c54.xml'

from mmda.parsers import PDFPlumberParser
from mmda.types import Document

pdf_plumber = PDFPlumberParser()
doc: Document = pdf_plumber.parse(input_pdf_path=PDF_PATH)
doc.fields

from mmda.parsers.grobid_augment_existing_document_parser import GrobidAugmentExistingDocumentParser
parser = GrobidAugmentExistingDocumentParser(config_path='../../src/mmda/parsers/grobid.config', check_server=False)

xml = open(XML_PATH).read()
doc = parser._parse_xml_onto_doc(xml, doc) # error happens here when trying to annotate sections onto the doc
```

regan-huff (approved):
This looks good to me! Happy for Kyle to weigh in also though

PR #278, opened 2023-09-02
bugfix for box_groups_to_span_groups when center=True

geli-gel:
this is a bug fix for an error made in https://github.com/allenai/mmda/pull/276
It's possible to find tokens from the original Document overlapping with the MergedSpan SpanGroup that are already allocated to a different SpanGroup, so we need to check first if it's in the list of available tokens before attempting to remove it. 

todo:
- [ ] publish new version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.9.13'
+version = '0.9.14'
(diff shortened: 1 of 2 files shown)


regan-huff (approved):
👍

regan-huff:
@geli-gel failed for me too, will try again in a bit

PR #283, opened 2024-06-05
Allows use of pydantic 2 by dependents.

cmwilhelm:
Some applications wants to use pydantic2
but have been blocked by strict constraint on
pydantic1 in mmda. This changeset allows the
basic library to be used with either major
version (but preserves the 1.x requirement
for specific models so as not to break anything
in S2's SPP pipeline).


geli-gel (approved):
👍🏽

PR #287, opened 2026-02-27
Update to supported python/CUDA versions

rodneykinney:
Update mmda services to be deployable again, hoping to fix https://github.com/allenai/scholar/issues/41682

Update python and CUDA to currently-supported versions https://github.com/allenai/scholar/issues/41718

Tests pass when run with timo code from https://github.com/allenai/timo/pull/359

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.9.25'
+version = '0.9.26'
 description = 'MMDA - multimodal document analysis'
 authors = [
     {name = 'Allen Institute for Artificial Intelligence', email = 'contact@allenai.org'},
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -118,11 +119,9 @@ bibentry_predictor_gpu = [
     'pydantic>=1,<2'
 ]
 bibentry_detection_predictor = [
-    'Pillow<10',
```
ca16, line 121: wondering why we're pinning less than before here?

rodneykinney: OMG, package version conflicts are everywhere, but Claude is amazing at figuring them out. IMO we should be pinning everything with a max version number

rodneykinney: I'm not sure why it decided to remove the pin. I would rather it replace with a larger upper-bound version number. I will instruct it to take a pass and put upper bounds everywhere

ca16 (approved):

PR #288, opened 2026-05-06
Account for layout parser files that have disappeared

ca16:
This PR adjusts the TIMO config for layout parser, to pull the files that layout parser used to find in dropbox from S3 instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = 'mmda'
-version = '0.9.18'
+version = '0.9.27'
 description = 'MMDA - multimodal document analysis'
 authors = [
     {name = 'Allen Institute for Artificial Intelligence', email = 'contact@allenai.org'},
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ heuristic_predictors = [
 ]
 lp_predictors = [
     'layoutparser',
+    'timm==0.9.12',
     'torch',
     'torchvision',
     'effdet',
--- a/src/ai2_internal/config.yaml
+++ b/src/ai2_internal/config.yaml
@@ -56,6 +56,9 @@ model_variants:
     # Whether this model supports CUDA GPU acceleration
     cuda: True
 
+    # Full S3 path to tar.gz'ed artifacts archive, nullable
+    artifacts_s3_path: s3://ai2-timo-registry/model-artifacts/layout-parser2/hopefully-what-was-in-dropbox-pulled-from-ecs-april-2026.tar.gz
+
     # Python path to a fn in <model_package_name>==<model_package_version> that
     # returns a unittest.TestCase. Builder function receives a model container
     # as its sole argument.
--- a/src/ai2_internal/layout_parser/interface.py
+++ b/src/ai2_internal/layout_parser/interface.py
@@ -8,6 +8,7 @@ as a definition of the objects it expects, and those it returns.
 
 import logging
 from typing import List
+from pathlib import Path
 
 import torch
 from pydantic import BaseModel, BaseSettings, Field
@@ -39,7 +40,10 @@ class PredictorConfig(BaseSettings):
     Configuration required by the model to do its work.
     Uninitialized fields will be set via Environment variables.
     """
-    weights_paths = ["lp://efficientdet/PubLayNet", "lp://efficientdet/MFD"]
+    weights_config_to_file = {
+        "lp://efficientdet/PubLayNet": "ukbw5s673633hsw-publaynet-tf_efficientdet_d0.pth.tar",
+        "lp://efficientdet/MFD": "dkr22iux7thlhel-mfd-tf_efficientdet_d0.pth.tar",
+    }
 
 class Predictor:
     """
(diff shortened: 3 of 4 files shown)


ca16:
Btw CI is also failing off main (with an extra space in the README): https://github.com/allenai/mmda/pull/289

ca16:
More on CI: I tried changing some things a bit in #289 . Here's a summary of what I found:

- After doing some pinning, lots of tests fail with the same error we started seeing after the dropbox files disappeared. This seems expected.
- Getting everything to work on python 3.9 seems to also require some code changes around torch expectations (see details below for the `build (3.9) job`). I didn't get all the way to actually making changes that would make this pass.
- Among other changes, test_word_predictors seemed to need python 3.9 to work (maybe there was another way to get there with 3.8, but I didn't explore that too much)

Since we're seeing these errors with the code in main, I don't think the changes in this PR are causing them. So I think the question that's left is if we want to try and fix anything those tests are pointing at as part of this PR:
- missing dropbox files - this is a real problem, so I think it might be good that tests are are catching this (or would be after some pinning). I'm inclined to do nothing further around this in this PR.
- making everything pass for python 3.8 and/or python 3.9 - I sort of lean towards not going down this path until we know someone is still using it/still needs it. And even then, maybe focusing on fixing just what is in use. For our purposes (getting layout parser working in our system again), we can construct a working environment with what's in this PR. So  I lean towards doing nothing further in this PR for this too.

Thoughts?

Details:

<details>

### build (3.8) job
At first, we saw errors like
```
FAILED tests/test_recipes/test_core_recipe.py::TestCoreRecipe::test_to_from_json - RuntimeError: weights_only=True blocked loading this checkpoint because it requires non-allowlisted pickle globals. No automatic unsafe pickle fallback is performed. If this checkpoint is trusted, retry with weights_only=False.
```
pinning timm in lp_predictors seemed to get us past this.

Then we saw errors like
```
FAILED tests/test_predictors/test_section_nesting_predictor.py::TestSectionNestingPredictor::test_finds_sections - _pickle.UnpicklingError: invalid load key, '<'.
```
These look like the error we're seeing because of the missing dropbox files. So these should be failing.


### test_vila_predictors job
At first we saw this
```
FAILED tests/test_predictors/test_vila_predictors.py::TestFigureVilaPredictors::test_vila_predictors - RuntimeError: weights_only=True blocked loading this checkpoint because it requires non-allowlisted pickle globals. No automatic unsafe pickle fallback is performed. If this checkpoint is trusted, retry with weights_only=False.
```
pinning tmm in vila_predictors seemed to get us past this.

Then we saw this
```
FAILED tests/test_predictors/test_vila_predictors.py::TestFigureVilaPredictors::test_vila_predictors - _pickle.UnpicklingError: invalid load key, '<'.
```
This looks like the error we're seeing because of the missing dropbox files. So this should be failing.

### test_word_predictors job

At first we saw errors like this:
```
  Building wheel for tokenizers (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building wheel for tokenizers (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [9 lines of output]
      Running `maturin pep517 build-wheel -i /opt/hostedtoolcache/Python/3.8.18/x64/bin/python --compatibility off`
      ⚠️  Warning: `project.version` field is required in pyproject.toml unless it is present in the `project.dynamic` list
      🍹 Building a mixed python/rust project
      🔗 Found pyo3 bindings with abi3 support
      🐍 Found CPython 3.8 at /opt/hostedtoolcache/Python/3.8.18/x64/bin/python
      📡 Using build options features, bindings from pyproject.toml
      💥 maturin failed
        Caused by: None of the found Python interpreters (CPython 3.8) are compatible with the abi3 minimum version (>= 3.9). Please install a compatible Python interpreter.
      Error: command ['maturin', 'pep517', 'build-wheel', '-i', '/opt/hostedtoolcache/Python/3.8.18/x64/bin/python', '--compatibility', 'off'] returned non-zero exit status 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tokenizers
Successfully built mmda ncls
Failed to build tokenizers
ERROR: Failed to build installable wheels for some pyproject.toml based projects (tokenizers)
Error: Process completed with exit code 1.
```
Updating which python version we use for this job (3.9 instead of 3.8) made this job start passing. Note: after a little more testing, found that pinning numpy (to 1.24.4, version from the surviving task) is also necessary. 

Otherwise (if we use python 3.9 but don't pin numpy) we see errors like:
```
ERROR tests/test_predictors/test_svm_word_predictor.py - ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
here too

### build (3.9) job
```
ERROR tests/test_utils/test_tools.py - ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
pinning numpy to 1.24.4 (from the surviving task) gets us past that
```
E                   _pickle.UnpicklingError: Weights only load failed. In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
```
pinning torch to what is in the surviving task (2.2.0) got us past these.

Then there are two kinds of errors:
```
E       _pickle.UnpicklingError: invalid load key, '<'.
```
These look like the error we're seeing because of the missing dropbox files. So these should be failing.

```
E           ValueError: Due to a serious vulnerability issue in `torch.load`, even with `weights_only=True`, we now require users to upgrade torch to at least v2.6 in order to use the function. This version restriction does not apply when loading files with safetensors.
```
Fixing this seems like a bigger issue.

</details>

rodneykinney (approved):

PR #289, opened 2026-05-06
[DO NOT MERGE] test CI

ca16:
Trying to see if CI fails off the code in main. (and then some testing to see what might fix it)

### DIFF
--- a/.github/workflows/mmda-ci.yml
+++ b/.github/workflows/mmda-ci.yml
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.8'
           architecture: x64
       - run: |
           sudo apt-get update
--- a/.github/workflows/mmda-ci.yml
+++ b/.github/workflows/mmda-ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.8"]
+        python-version: ["3.9"]
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/mmda-ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.8'
+          python-version: '3.9'
```
ca16: This makes the test_word_predictors CI job pass. claude code says it was failing because 'pip can't find a pre-built wheel for your CI's Python 3.8'. And 3.9 is one of the versions we look at in the build CI job above so bumping here seems sort of okay?

pyproject.toml
```diff
@@ -89,6 +90,7 @@ hf_predictors = [
     'smashed==0.1.10',
 ]
 vila_predictors = [
+    'timm==0.9.12',
```
ca16: This gets the test_vila_predictors CI job to fail with an error similar to the one we were hitting after the dropbox files disappeared.

pyproject.toml
```diff
@@ -125,14 +127,14 @@ bibentry_detection_predictor = [
     'pydantic<=1,<2',
 ]
 citation_links = [
-    'numpy',
+    'numpy==1.24.4',
```
ca16: This gets us past errors in the build (3.9) job that look like this:
```
ERROR tests/test_utils/test_tools.py - ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
claude code suggested it was a numpy version issue, and I picked this version because it's what's in the surviving task

